PR #3, opened 2020-10-11
feat: Added CI

ZeLarpMaster:


Blond11516 (commented):
This may be better discussed in an issue, but what do you think of adding credo to the project/CI?

ZeLarpMaster:
Could indeed be a good idea, but would need to be a separate issue/PR since it's most likely gonna include some changes to the code

Blond11516:
Yeah good point. I'm gonna open an issue.

Blond11516 (approved):

PR #5, opened 2020-10-12
feat: Added Credo

### LINKED ISSUE #4: Consider using Credo
I think adding [Credo](https://hexdocs.pm/credo/overview.html) to the project would help to maintain a minimal amount of code quality.

It would obviously need to be validated by the CI.

ZeLarpMaster:
Fixes #4 

There's currently a bunch of warnings and things to fix, but I want to get opinions on the config before fixing any


Blond11516:
Did you stick with the default config? If not what did you change (and why)?

ZeLarpMaster:
Default config + these in the `Controversial and experimental checks` section
```elixir
{Credo.Check.Readability.StrictModuleLayout, [order: [:shortdoc, :moduledoc, :use, :import, :alias, :require, :behaviour]]},
{Credo.Check.Readability.MultiAlias, []},
{Credo.Check.Readability.SinglePipe, []},
{Credo.Check.Refactor.PipeChainStart, []}
```
`order:` param on `StrictModuleLayout` is default except `:behaviour` being put under `:require` instead of under `:moduledoc`. This one is because `alias` is useful when defining the behaviour so we don't end up doing `@behaviour Mobius.Foo.Bar` with an `alias Mobius.Foo.Bar` right under it

`SinglePipe` prevents doing single pipes like `foo |> bar()` so pipes can only be used with multiple ones

`MultiAlias` is to enforce `alias Foo.Bar\n alias Foo.Baz` instead of `alias Foo.{Bar, Baz}` which is, in my opinion, more readable and easier to deal with when refactoring (cause you can Find & Replace globally for a module name that you'd move)

`PipeChainStart` is to prevent a function call at the top of pipes and enforce a variable being used. This one I'm not so sure about, but I feel like it could be a good idea

Blond11516:
Those all sound good to me! I think we can go ahead with this config.

Blond11516 (approved):

PR #7, opened 2020-10-25
feat: Adding tests for services

ZeLarpMaster:
I know this is a lot of changes, so here's a breakdown of what changed

This includes a bunch of refactoring in tests. Utility functions were added to avoid copying the same lines over and over. The socket stub now no longer stores all messages received but instead sends them back to the test pid for a simpler and more deterministic way of testing (by using `assert_receive` and `assert_received` which are much easier to use and debug)

The tests I added also made me discover a handful of bugs:

- Heartbeat and Shard had their `restart` set to `:permanent` which meant they'd be restarted everytime they exit regardless of the reason. This is not what we want in some cases (for example, when we shutdown the Heartbeat, we expect Shard to eventually start it back up). So they've been changed to `:transient` which makes the supervisor restart them only if the reason is anything other than `:normal`, `:shutdown` or `{:shutdown, term}`. The few times where we'd make those services stop themselves have been modified so their exit reason fits this.
- Shard didn't update its internal sequence number at all. This has been fixed with `update_seq` alongside a tiny refactor to pipe the state into `update_seq` and `update_state_by_event`
- The socket stub wasn't replicating the Socket behaviour as expected (namely notifying for `down` and `up` properly) which lead to the heartbeat service not being cleaned up which in turn caused errors.

### DIFF
--- a/test/mobius/services/heartbeat_test.exs
+++ b/test/mobius/services/heartbeat_test.exs
@@ -1,0 +1,33 @@
+defmodule Mobius.Services.HeartbeatTest do
+  use ExUnit.Case
+
+  import Mobius.Fixtures
+
+  alias Mobius.Core.Opcode
+  alias Mobius.Core.ShardInfo
+  alias Mobius.Services.Socket
+  alias Mobius.Stubs
+
+  setup :reset_services
+  setup :stub_socket
+
+  @shard ShardInfo.new(number: 0, count: 1)
+
+  test "sends heartbeat regularly", ctx do
+    data = %{d: %{heartbeat_interval: 2_000}, op: Opcode.name_to_opcode(:hello), t: nil, s: nil}
+    Socket.notify_payload(data, @shard)
+
+    ctx.socket
+    |> Stubs.Socket.has_message?(fn msg ->
+      msg == Opcode.heartbeat(0)
+    end)
+    |> assert
+
+    data = %{d: nil, t: nil, s: nil, op: Opcode.name_to_opcode(:heartbeat_ack)}
+    Socket.notify_payload(data, @shard)
+  end
+
+  test "sends heartbeat immediately if requested"
+  test "closes the socket if no ack since last heartbeat"
+  test "updates ping when receives an ack"
+end
--- a/test/mobius/services/heartbeat_test.exs
+++ b/test/mobius/services/heartbeat_test.exs
@@ -4,13 +4,11 @@ defmodule Mobius.Services.HeartbeatTest do
   import Mobius.Fixtures
 
   alias Mobius.Core.Opcode
-  alias Mobius.Core.ShardInfo
   alias Mobius.Services.Heartbeat
 
+  setup :get_shard
   setup :reset_services
   setup :stub_socket
-
-  @shard ShardInfo.new(number: 0, count: 1)
 
   test "sends heartbeat regularly" do
     send_hello(500)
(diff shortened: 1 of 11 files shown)


ZeLarpMaster:
Oh and also, coverage is now ~90% from ~60%

test/mobius/services/heartbeat_test.exs
```diff
@@ -0,0 +1,57 @@
...
+  import Mobius.Fixtures
+
+  alias Mobius.Core.Opcode
+  alias Mobius.Core.ShardInfo
+  alias Mobius.Services.Heartbeat
+
+  setup :reset_services
+  setup :stub_socket
+
+  @shard ShardInfo.new(number: 0, count: 1)
```
Blond11516: nit: Does the shard need to be global if it's only used in one test?

test/mobius/services/shard_test.exs
```diff
@@ -0,0 +1,78 @@
...
+  setup :reset_services
+  setup :stub_socket
+  setup :handshake_shard
+
+  @shard ShardInfo.new(number: 0, count: 1)
+
+  test "attempts to resume if socket closes with resumable code", ctx do
+    send_payload(op: :dispatch, type: :TYPING_START, seq: 2)
+
+    socket_closed_by_server(1001, "Going away")
```
Blond11516: I'd extract the code to a variable so you don't have to know what the code means to understand the test. (Same with 4009 and 4013)

test/mobius/services/shard_test.exs
```diff
@@ -0,0 +1,78 @@
...
+    send_payload(op: :invalid_session, data: true)
+
+    send_hello()
+
+    resume =
+      ctx.token
+      |> Gateway.new()
+      |> Gateway.update_seq(1)
+      |> Gateway.set_session_id(ctx.session_id)
+      |> Opcode.resume()
```
Blond11516: Maybe this whole pipe could be extracted to a function? It would avoid repeating it twice and a good name would make it's role more clear.

test/support/fixtures.ex
```diff
@@ -48,6 +44,31 @@ defmodule Mobius.Fixtures do
...
+      op: Opcode.name_to_opcode(Keyword.fetch!(opts, :op)),
+      d: Keyword.get(opts, :data),
+      t: Keyword.get(opts, :type),
+      s: Keyword.get(opts, :seq)
+    }
+
+    Socket.notify_payload(data, @shard)
+  end
+
+  def socket_closed_by_server(close_num, reason) do
```
Blond11516: The name feels a little confusing. The current name kinda makes me think that the function should return a boolean, but it's really more of an action.

test/support/socket_stub.ex
```diff
@@ -63,56 +52,46 @@ defmodule Mobius.Stubs.Socket do
...
+      test_pid: nil
     }
 
     {:ok, state}
   end
 
   @impl GenServer
   def handle_cast({:send, payload}, state) do
-    {:noreply, update_in(state.messages, &(&1 ++ [{:msg, payload}]))}
+    unless state.test_pid == nil do
```
Blond11516:
```suggestion
    if state.test_pid != nil do
```

This is really personal, but I'm not a big fan of `unless`. I guess it would be alright for cases where you have to do `if not some_boolean_function()` but here I have to stop for a second to make sure I get what's going on.

Blond11516 (dismissed):

test/mobius/services/heartbeat_test.exs
ZeLarpMaster: It was used in like 4 tests before, but after refactoring not anymore. Since I've been copy pasting this line everywhere I should probably put it in the `ctx` through fixtures 🤔

test/mobius/services/shard_test.exs
ZeLarpMaster: Well `1001` literally means `Going away`. I copy pasted the name of the code as the reason. I'm not sure what information I could add to make it clearer?

test/support/fixtures.ex
ZeLarpMaster: `close_socket_from_server`? I see what you mean I just couldn't find a better name when I wrote this at 3am

Blond11516 (approved):

PR #12, opened 2020-11-27
feat: Store shard infos on ETS

ZeLarpMaster:
The ETSShelf's only purpose is holding onto ets tables when their owner dies. This prevents data loss when processes which originally held on the table are restarted

This removes a bottleneck from the Bot service just for listing shards and will help keep some features concurrent (ie.: sending a message to a service on all shards won't need to go through the Bot first)

### DIFF
--- a/lib/mobius/core/shard_list.ex
+++ b/lib/mobius/core/shard_list.ex
@@ -1,0 +1,35 @@
+defmodule Mobius.Core.ShardList do
+  @moduledoc false
+
+  alias Mobius.Core.ShardInfo
+
+  @spec table_options() :: [atom]
+  def table_options do
+    [:ordered_set, :protected]
+  end
+
+  @spec list_shards(:ets.tab()) :: [ShardInfo.t()]
+  def list_shards(table) do
+    shards = :ets.match(table, {:"$1", :_})
+    List.flatten(shards)
+  end
+
+  @spec are_all_shards_ready?(:ets.tab()) :: boolean
+  def are_all_shards_ready?(table) do
+    # :ets.fun2ms(fn {_, state} when state != :ready -> state end)
+    spec = [{{:_, :"$1"}, [{:"/=", :"$1", :ready}], [:"$1"]}]
+    :ets.select(table, spec) == []
+  end
+
+  @spec add_shard(:ets.tab(), ShardInfo.t()) :: ShardInfo.t()
+  def add_shard(table, %ShardInfo{} = shard) do
+    :ets.insert(table, {shard, :starting})
+    shard
+  end
+
+  @spec update_shard_ready(:ets.tab(), ShardInfo.t()) :: ShardInfo.t()
+  def update_shard_ready(table, %ShardInfo{} = shard) do
+    :ets.insert(table, {shard, :ready})
+    shard
+  end
+end
--- a/lib/mobius/core/shard_list.ex
+++ b/lib/mobius/core/shard_list.ex
@@ -19,6 +19,7 @@ defmodule Mobius.Core.ShardList do
   @doc "True if all shards have state `:ready`"
   @spec are_all_shards_ready?(:ets.tab()) :: boolean
   def are_all_shards_ready?(table) do
+    # The match spec was generated with the following line in iex:
     # :ets.fun2ms(fn {_, state} when state != :ready -> state end)
     spec = [{{:_, :"$1"}, [{:"/=", :"$1", :ready}], [:"$1"]}]
     :ets.select(table, spec) == []
--- a/lib/mobius/core/shard_list.ex
+++ b/lib/mobius/core/shard_list.ex
@@ -3,17 +3,20 @@ defmodule Mobius.Core.ShardList do
 
   alias Mobius.Core.ShardInfo
 
+  @doc "Options for the ETS table creation"
   @spec table_options() :: [atom]
   def table_options do
     [:ordered_set, :protected]
   end
 
+  @doc "Lists all shards regardless of state"
   @spec list_shards(:ets.tab()) :: [ShardInfo.t()]
   def list_shards(table) do
     shards = :ets.match(table, {:"$1", :_})
     List.flatten(shards)
   end
 
+  @doc "True if all shards have state `:ready`"
   @spec are_all_shards_ready?(:ets.tab()) :: boolean
   def are_all_shards_ready?(table) do
     # :ets.fun2ms(fn {_, state} when state != :ready -> state end)
@@ -21,12 +24,14 @@ defmodule Mobius.Core.ShardList do
     :ets.select(table, spec) == []
   end
 
+  @doc "Adds a shard to the list with state `:starting`"
   @spec add_shard(:ets.tab(), ShardInfo.t()) :: ShardInfo.t()
   def add_shard(table, %ShardInfo{} = shard) do
     :ets.insert(table, {shard, :starting})
     shard
   end
 
+  @doc "Updates a shard's state to `:ready`"
   @spec update_shard_ready(:ets.tab(), ShardInfo.t()) :: ShardInfo.t()
   def update_shard_ready(table, %ShardInfo{} = shard) do
     :ets.insert(table, {shard, :ready})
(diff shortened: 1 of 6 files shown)


lib/mobius/services/ets_shelf.ex
```diff
@@ -0,0 +1,59 @@
...
+      ArgumentError -> {:reply, {:error, :ets_error}, state}
+    end
+  end
+
+  @impl GenServer
+  def handle_info({:"ETS-TRANSFER", table, _pid, _data}, state) do
+    {:noreply, Map.put(state, table, nil)}
+  end
+
+  defp create_or_give(name, pid, opts, state) do
```
Blond11516, line 49: Shouldn't this be called `create_and_give`? Creating the table doesn't exclude the action of giving it away.

lib/mobius/services/ets_shelf.ex
```diff
@@ -0,0 +1,59 @@
...
+  @timeout 1_000
+
+  @spec start_link(keyword) :: GenServer.on_start()
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @spec create_table(atom, [atom | tuple]) :: :ok | {:error, :ets_error}
+  def create_table(name, opts) when is_atom(name) do
+    # Creates the table or gives it back if the server already owns it
```
Blond11516:
```suggestion
    # Creates the table or gives it back if the shelf already owns it
```
Server is a pretty vague term

lib/mobius/services/ets_shelf.ex
```diff
@@ -0,0 +1,59 @@
...
+
+  @impl GenServer
+  @spec init(keyword) :: {:ok, map}
+  def init(_opts) do
+    {:ok, %{}}
+  end
+
+  @impl GenServer
+  def handle_call({:create, name, pid, opts}, _from, state) do
+    # Set the server as the heir
```
Blond11516:
```suggestion
    # Set the shelf as the heir
```

test/mobius/services/ets_shelf_test.exs
```diff
@@ -0,0 +1,38 @@
...
+  setup do
+    [table: :"test_table_#{random_hex(8)}"]
+  end
+
+  test "create_table/2 returns {:error, :ets_error} if table already exists", ctx do
+    table = :ets.new(ctx.table, [:named_table])
+    assert {:error, :ets_error} == Shelf.create_table(table, [])
+  end
+
+  test "create_table/2", ctx do
```
Blond11516: You could give a more explicit test description. Maybe something like "create_table/2 inherits table when owner dies"?

test/mobius/services/ets_shelf_test.exs
```diff
@@ -0,0 +1,38 @@
...
+    # Table is owned by the task
+    assert {:error, :ets_error} == Shelf.create_table(ctx.table, [])
+    Task.shutdown(task)
+
+    # Table went back to the shelf
+    assert :ok == Shelf.create_table(ctx.table, [])
+
+    assert :ets.insert(ctx.table, {123, :my_thing})
+  end
+end
```
Blond11516, line 40: Have you considered using `describe` to group tests for the same function?

lib/mobius/services/bot.ex
```diff
@@ -5,14 +5,15 @@ defmodule Mobius.Services.Bot do
```
Blond11516, line 5: What would you think about extracting `:ets` calls to another module (could be done in a future PR)? I think the bot has more important things to worry about than the details of interacting with an ETS.

Blond11516 (dismissed):

lib/mobius/services/ets_shelf.ex
ZeLarpMaster: If the table exists, the shelf gives it away without doing anything else. If the table doesn't exist, the shelf creates it and then gives it away. But you can't create it if it already exists. Neither "or" or "and" really mean the right thing here. It can't be created if it exists so "and" is false. "or" can either be mutually exclusive in which case it's false because it does both when it creates, or inclusive in which case it's technically true because it either gives or creates and gives.

lib/mobius/services/bot.ex
ZeLarpMaster: It probably should, but I'm not sure of where it would go. Would it be a part of core even though it has global side effects? It can't really because it's also using `ETSShelf` which is a service and core should absolutely never be using services. Would it be a part of a different service? That kinda feels weird. Maybe we leave the ETSShelf usage to Bot and put everything else in a Core module? How does that sound?

lib/mobius/services/ets_shelf.ex
Blond11516: Yeah, it's quite a weird one to name indeed. It's not a big issue, but maybe a more accurate solution would be to have a `get_or_create` function piped into a `give` function (oh straight into `:ets.give_away` at this point).

But it's fine like that. Let's change it later if we still feed the need to.

test/mobius/services/ets_shelf_test.exs
ZeLarpMaster: I could be using a describe, but I feel like it would be redundant with the test file itself since the ETSShelf doesn't have any other public function. Should I still use a describe here?

test/mobius/services/ets_shelf_test.exs
ZeLarpMaster: I think that's a bit too "whitebox-y" for me, I went with "gives the table after original owner dies" instead

lib/mobius/services/bot.ex
Blond11516: Yeah I agree that Core isn't the right place. I like your suggestion, although for a simpler solution making a "Shard" service wouldn't feel that weird to me. That said I just got started reading the bee book so the exact role of a service is still a little unclear to me. I'll leave the final decision up to you.

test/mobius/services/ets_shelf_test.exs
Blond11516: I'd use it since both test names start with the same function name, but it really is my personal opinion. No idea what the Elixir convention for this is (if there is any) 🤷

Blond11516 (dismissed):

test/mobius/services/ets_shelf_test.exs
ZeLarpMaster: I don't think there's any so I'll just add the describe

lib/mobius/services/bot.ex
ZeLarpMaster: Well the Shard service already exists and does a completely different thing. I think I'll go with the core module idea of keeping the ETSShelf part in Bot and abstracting the `:ets` calls into it. It also makes usage easier for the other modules which want to use the ets table. I'll change it on this PR

lib/mobius/core/shard_list.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  @spec list_shards(:ets.tab()) :: [ShardInfo.t()]
+  def list_shards(table) do
+    shards = :ets.match(table, {:"$1", :_})
+    List.flatten(shards)
+  end
+
+  @spec are_all_shards_ready?(:ets.tab()) :: boolean
+  def are_all_shards_ready?(table) do
+    # :ets.fun2ms(fn {_, state} when state != :ready -> state end)
```
Blond11516, line 23: ?

Blond11516 (dismissed):

lib/mobius/core/shard_list.ex
ZeLarpMaster: `:ets.select`'s match specification tends to be very hard to read so there's a utility named `:ets.fun2ms` which converts an arbitrary function into a match specification. The problem is that function can't be used in code (neither runtime or build-time work). However it does work in iEx so I write the `:ets.fun2ms` version in a comment as documentation for the reader and I use its return value where I needed it.

Would it clarify if I added `# The spec was generated with the following line in iex:` before each of these?

ZeLarpMaster: Added the other comment to give a bit of context on this one

Blond11516 (approved):

PR #13, opened 2020-12-04
feat: Discord event subscription POC

ZeLarpMaster:
Currently is a POC of how the data flow of events will work

TODO:
- [x] Call a Core module which does the mapping (can be identity for now)
- [x] Add event name validation
- [x] Write documentation in `Mobius.Actions.Events`
- [x] Send `READY` from `Mobius.Services.Bot` when all shards are ready
- [x] Write tests for `Mobius.Core.Event`
- [x] Write tests for `Mobius.Actions.Events`
- [x] Write tests for `Mobius.Services.EventPipeline`
- [x] Write test checking `Mobius.Services.Shard` is sending the events

### DIFF
--- a/lib/mobius/actions/events.ex
+++ b/lib/mobius/actions/events.ex
@@ -3,6 +3,7 @@ defmodule Mobius.Actions.Events do
   Functions related to events
   """
 
+  alias Mobius.Core.Event
   alias Mobius.Services.EventPipeline
 
   @doc """
@@ -18,7 +19,7 @@ defmodule Mobius.Actions.Events do
   The caller process is then sent events whose name is in the set
   as messages in the format `{event_name, event_data}`
   where `event_name` is the name of the event as an UPPER_CASE atom
-  such as `:CHANNEL_CREATE` and where `event_data` is a struct
+  such as `:CHANNEL_CREATE` and `event_data` is a struct
   containing all the information given by the event
 
   If the set is not empty,
@@ -27,11 +28,11 @@ defmodule Mobius.Actions.Events do
   See https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events
   for a list of events and their associated data
   """
-  @spec subscribe([atom]) :: :ok
+  @spec subscribe([Event.names()]) :: :ok
   def subscribe(events \\ []) do
     # TODO: Validate event names
     # TODO: Validate intents
-    EventPipeline.subscribe_event_categories(events)
+    EventPipeline.subscribe(events)
   end
 
   @doc """
--- a/lib/mobius/actions/events.ex
+++ b/lib/mobius/actions/events.ex
@@ -5,13 +5,41 @@ defmodule Mobius.Actions.Events do
 
   alias Mobius.Services.EventPipeline
 
+  @doc """
+  Subscribes the calling process to a set of (or all) events
+
+  Subscribing multiple times with overlapping events will result
+  in duplicate events. As such, it is discouraged to subscribe
+  multiple times from the same process since this may cause
+  performance degradation
+
+  If an empty list (the default) is given, subscribes to all events
+
+  The caller process is then sent events whose name is in the set
+  as messages in the format `{event_name, event_data}`
+  where `event_name` is the name of the event as an UPPER_CASE atom
+  such as `:CHANNEL_CREATE` and where `event_data` is a struct
+  containing all the information given by the event
+
+  If the set is not empty,
+  the event will be sent if `Enum.member?(events, event_name)` is true
+
+  See https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events
+  for a list of events and their associated data
+  """
   @spec subscribe([atom]) :: :ok
-  def subscribe(events) do
+  def subscribe(events \\ []) do
     # TODO: Validate event names
     # TODO: Validate intents
     EventPipeline.subscribe_event_categories(events)
   end
 
+  @doc """
+  Unsubscribes the calling process from all events it was subscribed to
+
+  If the calling process had `subscribe/1`d multiple times,
+  this unsubscribes from all of them
+  """
   @spec unsubscribe() :: :ok
   def unsubscribe do
     EventPipeline.unsubscribe()
--- a/lib/mobius/actions/events.ex
+++ b/lib/mobius/actions/events.ex
@@ -5,6 +5,8 @@ defmodule Mobius.Actions.Events do
 
   alias Mobius.Core.Event
   alias Mobius.Services.EventPipeline
+  alias Mobius.Validations.EventValidator
+  alias Mobius.Validations.Utils
 
   @doc """
   Subscribes the calling process to a set of (or all) events
@@ -28,11 +30,12 @@ defmodule Mobius.Actions.Events do
   See https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events
   for a list of events and their associated data
   """
-  @spec subscribe([Event.names()]) :: :ok
+  @spec subscribe([Event.names()]) :: Utils.output()
   def subscribe(events \\ []) do
-    # TODO: Validate event names
-    # TODO: Validate intents
-    EventPipeline.subscribe(events)
+    with :ok <- EventValidator.validate_names(events) do
+      # TODO: Validate intents
+      EventPipeline.subscribe(events)
+    end
   end
 
   @doc """
--- a/lib/mobius/actions/events.ex
+++ b/lib/mobius/actions/events.ex
@@ -3,13 +3,14 @@ defmodule Mobius.Actions.Events do
   Functions related to events
   """
 
-  alias Mobius.Core.Event
   alias Mobius.Services.EventPipeline
   alias Mobius.Validations.EventValidator
   alias Mobius.Validations.Utils
 
   @doc """
   Subscribes the calling process to a set of (or all) events
+
+  Passing invalid event names will cause this function to return `{:errors, [String.t()]}`
 
   Subscribing multiple times with overlapping events will result
   in duplicate events. As such, it is discouraged to subscribe
@@ -30,7 +31,7 @@ defmodule Mobius.Actions.Events do
   See https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events
   for a list of events and their associated data
   """
-  @spec subscribe([Event.name()]) :: Utils.output()
+  @spec subscribe([any]) :: Utils.output()
   def subscribe(events \\ []) do
     with :ok <- EventValidator.validate_names(events) do
       # TODO: Validate intents
--- a/lib/mobius/actions/events.ex
+++ b/lib/mobius/actions/events.ex
@@ -30,7 +30,7 @@ defmodule Mobius.Actions.Events do
   See https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events
   for a list of events and their associated data
   """
-  @spec subscribe([Event.names()]) :: Utils.output()
+  @spec subscribe([Event.name()]) :: Utils.output()
   def subscribe(events \\ []) do
     with :ok <- EventValidator.validate_names(events) do
       # TODO: Validate intents
@@ -41,7 +41,7 @@ defmodule Mobius.Actions.Events do
   @doc """
   Unsubscribes the calling process from all events it was subscribed to
 
-  If the calling process had `subscribe/1`d multiple times,
+  If the calling process had called `subscribe/1` multiple times,
   this unsubscribes from all of them
   """
   @spec unsubscribe() :: :ok
--- a/lib/mobius/actions/events.ex
+++ b/lib/mobius/actions/events.ex
@@ -1,0 +1,19 @@
+defmodule Mobius.Actions.Events do
+  @moduledoc """
+  Functions related to events
+  """
+
+  alias Mobius.Services.EventPipeline
+
+  @spec subscribe([atom]) :: :ok
+  def subscribe(events) do
+    # TODO: Validate event names
+    # TODO: Validate intents
+    EventPipeline.subscribe_event_categories(events)
+  end
+
+  @spec unsubscribe() :: :ok
+  def unsubscribe do
+    EventPipeline.unsubscribe()
+  end
+end
--- a/lib/mobius/core/event.ex
+++ b/lib/mobius/core/event.ex
@@ -2,40 +2,41 @@ defmodule Mobius.Core.Event do
   @moduledoc false
 
   @valid_names [
-    "CHANNEL_CREATE",
-    "CHANNEL_UPDATE",
-    "CHANNEL_DELETE",
-    "CHANNEL_PINS_UPDATE",
-    "GUILD_CREATE",
-    "GUILD_UPDATE",
-    "GUILD_DELETE",
-    "GUILD_BAN_ADD",
-    "GUILD_BAN_REMOVE",
-    "GUILD_EMOJIS_UPDATE",
-    "GUILD_INTEGRATIONS_UPDATE",
-    "GUILD_MEMBER_ADD",
-    "GUILD_MEMBER_REMOVE",
-    "GUILD_MEMBER_UPDATE",
-    "GUILD_ROLE_CREATE",
-    "GUILD_ROLE_UPDATE",
-    "GUILD_ROLE_DELETE",
-    "INVITE_CREATE",
-    "INVITE_DELETE",
-    "MESSAGE_CREATE",
-    "MESSAGE_UPDATE",
-    "MESSAGE_DELETE",
-    "MESSAGE_DELETE_BULK",
-    "MESSAGE_REACTION_ADD",
-    "MESSAGE_REACTION_REMOVE",
-    "MESSAGE_REACTION_REMOVE_ALL",
-    "MESSAGE_REACTION_REMOVE_EMOJI",
-    "PRESENCE_UPDATE",
-    "TYPING_START",
-    "USER_UPDATE",
-    "VOICE_STATE_UPDATE",
-    "VOICE_SERVER_UPDATE",
-    "WEBHOOKS_UPDATE"
+    :CHANNEL_CREATE,
+    :CHANNEL_UPDATE,
+    :CHANNEL_DELETE,
+    :CHANNEL_PINS_UPDATE,
+    :GUILD_CREATE,
+    :GUILD_UPDATE,
+    :GUILD_DELETE,
+    :GUILD_BAN_ADD,
+    :GUILD_BAN_REMOVE,
+    :GUILD_EMOJIS_UPDATE,
+    :GUILD_INTEGRATIONS_UPDATE,
+    :GUILD_MEMBER_ADD,
+    :GUILD_MEMBER_REMOVE,
+    :GUILD_MEMBER_UPDATE,
+    :GUILD_ROLE_CREATE,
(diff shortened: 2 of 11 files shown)


lib/mobius/services/event_pipeline.ex
```diff
@@ -0,0 +1,35 @@
...
+    Task.Supervisor.child_spec(name: __MODULE__)
+  end
+
+  @doc "Sends an event to all subscribed processes"
+  @spec notify_event(atom, any) :: :ok
+  def notify_event(name, data) do
+    Task.Supervisor.start_child(__MODULE__, fn ->
+      # TODO: Map data depending on name
+      PubSub.publish(@pubsub_topic, name, data)
+    end)
```
Blond11516: Why are we doing this a task?

lib/mobius/services/shard.ex
```diff
@@ -176,13 +178,19 @@ defmodule Mobius.Services.Shard do
...
 
   defp update_state_by_event(state, %{t: :READY, d: d}) do
-    # TODO: Use a PubSub instead of this
     Bot.notify_ready(state.shard)
     set_session(state, d.session_id)
   end
 
   defp update_state_by_event(state, _payload), do: state
 
+  defp broadcast_event(state, %{t: type}) when type in [:READY], do: state
```
Blond11516: why not `type == :READY`? Future proofing?

Blond11516 (commented):
So, I have very little experience with PubSub, but if I understand correctly right now you can subscribe and unsubscribe to events, and every time an event comes in from the shard it's gonna be broadcasted, but there's no way for the subscriber to actually handle the event? Is that a planned feature or are you intending the user of `Mobius.Actions.Events` to manually handle the reception of the event?

ZeLarpMaster:
> there's no way for the subscriber to actually handle the event? Is that a planned feature or are you intending the user of `Mobius.Actions.Events` to manually handle the reception of the event?

`Mobius.Actions` stuff is intended as a "library" way of using Mobius. So for people who do use Mobius as a library and not a framework, they will indeed need to handle the messages themselves. The format of those messages is gonna be part of the documentation seen on hexdocs so I think it's gonna be fine. For people who use Mobius as a framework (aka through cogs), they'll have macros to help them with this.

Did you have a suggestion for another way to do this? I'm not sure how else this could work since the process subscribing could easily be a genserver which makes it hard to do anything to help them I think

lib/mobius/services/event_pipeline.ex
ZeLarpMaster: 2 reasons: 1. we don't want Mobius.Services.Shard to be spending its time mapping and publishing events. It has much more important work to do: keeping the bot alive. 2. we want to parallelize the work of mapping and publishing data. Using tasks gives us both. The reason why it's unlinked (through Task.Supervisor.start_child/2) is because we don't want mapping errors to bubble up and crash the Shard service. Instead we want to at most log a warning so we can fix the problem and ignore the event.

ZeLarpMaster: I'll add comments about that in the code

lib/mobius/services/shard.ex
ZeLarpMaster: Original intent was future proofing, I know for a fact there's gonna be at least one more event to not broadcast: [Guild Members Chunk](https://discord.com/developers/docs/topics/gateway#guild-members-chunk)

lib/mobius/core/event.ex
```diff
@@ -0,0 +1,82 @@
...
+          | :PRESENCE_UPDATE
+          | :TYPING_START
+          | :USER_UPDATE
+          | :VOICE_STATE_UPDATE
+          | :VOICE_SERVER_UPDATE
+          | :WEBHOOKS_UPDATE
+
+  @doc "Converts string event names to atoms if it's valid otherwise returns nil"
+  @spec parse_name(String.t()) :: names() | nil
+  def parse_name(name) when name in @valid_names, do: String.to_atom(name)
```
Blond11516:
```suggestion
  def parse_name(name) when name in @valid_names, do: String.to_existing_atom(name)
```

Blond11516:
> Did you have a suggestion for another way to do this? I'm not sure how else this could work since the process subscribing could easily be a genserver which makes it hard to do anything to help them I think

I'm just gonna spit out ideas. Feel free to tell me if they don't make sense.

There are two common use cases to handle here: the subscribing process could be a GenServer, or it could not be.

If it's not a GenServer, we could provide a function like `handle_event(event_handler)` (where event_handler is a function) that calls `receive` maybe does some processing on the event if required, and passes the event over the `event_handler`. This assumes the following:
* The subscriber has a receive loop (fairly safe assumption I think)
* The subscriber only expects to receive discord events as messages (one possible workaround for this would be to provide `handle_event` with a second function parameter for handling non-discord-event messages)

If the subscribing process is a GenServer, we could provide a macro that builds a `handle info` implementation with the required pattern matching/guard clauses for handling only discord events.

lib/mobius/core/event.ex
ZeLarpMaster: I tried that at first, but it doesn't work because the atoms aren't guaranteed to exist right away. It also doesn't really matter that much on a security stand point because the string has to be in the `@valid_names` list anyway

lib/mobius/actions/events.ex
```diff
@@ -0,0 +1,47 @@
...
+  multiple times from the same process since this may cause
+  performance degradation
+
+  If an empty list (the default) is given, subscribes to all events
+
+  The caller process is then sent events whose name is in the set
+  as messages in the format `{event_name, event_data}`
+  where `event_name` is the name of the event as an UPPER_CASE atom
+  such as `:CHANNEL_CREATE` and where `event_data` is a struct
+  containing all the information given by the event
```
Blond11516:
```suggestion
  The caller process is then sent events whose name is in the set
  as messages in the format `{event_name, event_data}`
  where `event_name` is the name of the event as an UPPER_CASE atom
  such as `:CHANNEL_CREATE` and `event_data` is a struct
  containing all the information given by the event
```

Also, do we actually need to send the event name outside of the struct? Isn't the struct going to contain this information anyway?

lib/mobius/services/event_pipeline.ex
```diff
@@ -0,0 +1,44 @@
...
+        parsed_data = Event.parse_data(parsed_name, data)
+        PubSub.publish(@pubsub_topic, parsed_name, parsed_data)
+      end
+    end)
+
+    :ok
+  end
+
+  @doc "Subscribes the calling process to an arbitrary set of events"
+  @spec subscribe_event_categories([atom]) :: :ok
```
Blond11516:
```suggestion
  @spec subscribe_event_categories([Event.names()]) :: :ok
```

lib/mobius/actions/events.ex
Blond11516: After reading the rest of the code, I see that the events don't actually contain a struct, but rather a map. Are there plans to create an actual struct?

lib/mobius/core/event.ex
Blond11516: That's a little unfortunate. What would you think of changing `@valid_names` to a list of atoms and convert them to strings when necessary? That way the atoms would always be defined and we could use `String.to_existing_atom`.

I'm also kind of assuming that we're going to need to validate event names as atoms at some point. The change does make a little less sense if that isn't true.

lib/mobius/services/event_pipeline.ex
Blond11516: Also, is "event_categories" the correct term? I feel like it's the only place where a notion of "categories" is mentioned.

ZeLarpMaster:
> we could provide a function like handle_event(event_handler)

I feel like messages are a part of Elixir APIs just as much as return types, so I'm not sure of why they need to be hidden behind functions/macros. Even if we do, it would need to be documented so the users aren't surprised to see an `{:event_name, data}` pop up in their other functions. In this case we wouldn't be hiding anything or providing anything more to the user which seems to defeat the purpose of a function.

> we could provide a macro that builds a handle info

We're most likely going to provide something like that in Cogs, but for those it's a little different since the user isn't really doing `use GenServer`. They're doing `use Cog` which then handles a bunch of stuff for the user as well as provide macros for binding to events.

The general goal of all `Mobius.Actions` modules is to provide a library feel where the user (potentially in a `.exs` script or even directly in their iex session) calls functions like `Mobius.Actions.Events.subscribe/1` and receives the events to either debug or do something quick. A secondary goal is to [provide all necessary APIs to the user so they can still use this lib without macros](https://hexdocs.pm/elixir/library-guidelines.html#avoid-macros).

I'm not sure if this all makes sense, but I hope it does

lib/mobius/core/event.ex
ZeLarpMaster: You're right that we'll need to validate event names as atoms (in this PR even). I'll try your suggestion and see if it works. I'm not sure of when the atom needs to exist in order to count as existing (does build time count?) which might be why it didn't work

lib/mobius/actions/events.ex
ZeLarpMaster: Yes, there's plans to translate the maps into structs. We could make a wrapper struct to contain both the event name and the data, but I don't really see an advantage to doing that when they still need to be split up for things like `subscribe/1` and parsing.

The event data's struct isn't unique per event name, for example all three of [`Channel Create`, `Channel Update`, and `Channel Delete` have a `Channel` struct as their data](https://discord.com/developers/docs/topics/gateway#channel-create). So the parsers are mapping [resources](https://discord.com/developers/docs/resources/channel#channel-object-channel-structure) and the event name is used to know which resource we received.

lib/mobius/services/event_pipeline.ex
ZeLarpMaster: I'm not really sure what to call it. Maybe I just keep it as `subscribe/1` and let the documentation explain its argument. How does that sound?

lib/mobius/core/event.ex
ZeLarpMaster: Looks like it worked! I am now completely confused by how `String.to_existing_atom/1` works, but I'll figure that out another day :P

lib/mobius/actions/events.ex
ZeLarpMaster: Also the parsing is gonna be reused for the HTTP responses, so adding a struct here would really only be a wrapper around `name` and `data`. Adding `name` inside the `data` wouldn't work for the HTTP responses and wouldn't work for nested structures (like users inside channels inside guilds)

lib/mobius/services/event_pipeline.ex
ZeLarpMaster: Actually it sounds much better so I'll just do that

lib/mobius/core/event.ex
Blond11516: If I'm not mistaken, the atoms only existed in the `@type` before right? Since typespecs aren't validated by Elixir itself I'd guess that all specs and types are removed as a first compilation pass, so the atoms were never compiled into the actual program.

Blond11516:
> I'm not sure if this all makes sense, but I hope it does

Yep, totally does. To be perfectly honest I had to stretch really hard to find that suggestion for non-GenServer support. I was really mostly trying to stir up conversation, but I agree with you view on this.

Blond11516:
This is looking good so far. I'll wait for the next update before doing another review.

lib/mobius/core/event.ex
ZeLarpMaster: I had them in `@valid_names`, but I guess since that was only used in a guard, the values weren't actually used whereas now it's also used in `is_valid_name/1` so that might be the difference

lib/mobius/actions/events.ex
```diff
@@ -0,0 +1,51 @@
...
+    with :ok <- EventValidator.validate_names(events) do
+      # TODO: Validate intents
+      EventPipeline.subscribe(events)
+    end
+  end
+
+  @doc """
+  Unsubscribes the calling process from all events it was subscribed to
+
+  If the calling process had `subscribe/1`d multiple times,
```
Blond11516:
```suggestion
  If the calling process had called `subscribe/1` multiple times,
```

Very nit, just feels a little weird imo :stuck_out_tongue:

lib/mobius/core/event.ex
```diff
@@ -0,0 +1,89 @@
+defmodule Mobius.Core.Event do
+  @moduledoc false
+
+  @valid_names [
```
Blond11516, line 4: Is there a specific you went with all caps names? I feel like lower case atoms would be more "conventional" for Elixir.

lib/mobius/core/event.ex
```diff
@@ -0,0 +1,89 @@
...
+          | :MESSAGE_REACTION_REMOVE_ALL
+          | :MESSAGE_REACTION_REMOVE_EMOJI
+          | :PRESENCE_UPDATE
+          | :TYPING_START
+          | :USER_UPDATE
+          | :VOICE_STATE_UPDATE
+          | :VOICE_SERVER_UPDATE
+          | :WEBHOOKS_UPDATE
+
+  @doc "Converts string event names to atoms if it's valid otherwise returns nil"
```
Blond11516:
```suggestion
  @doc "Converts string event names to atoms if it's valid. Returns `nil` otherwise."
```

lib/mobius/core/event.ex
```diff
@@ -0,0 +1,89 @@
...
+    :PRESENCE_UPDATE,
+    :TYPING_START,
+    :USER_UPDATE,
+    :VOICE_STATE_UPDATE,
+    :VOICE_SERVER_UPDATE,
+    :WEBHOOKS_UPDATE
+  ]
+  @valid_string_names Enum.map(@valid_names, &Atom.to_string/1)
+
+  @type names ::
```
Blond11516:
```suggestion
  @type name ::
```
(nit) Singular feels more appropriate for types.

lib/mobius/core/event.ex
```diff
@@ -0,0 +1,89 @@
...
+  @spec parse_name(String.t()) :: names() | nil
+  def parse_name(name) when name in @valid_string_names, do: String.to_existing_atom(name)
+  def parse_name(_name), do: nil
+
+  @doc "Returns true if the given name is a valid event name"
+  @spec is_event_name?(atom) :: boolean
+  def is_event_name?(name), do: name in @valid_names
+
+  @spec parse_data(names(), any) :: any
+  def parse_data(name, data) when name in @valid_names, do: data
```
Blond11516, line 100: This will crash if the name is not valid. Is this the intended behaviour?

Also this function could use a short doc, especially if crashing for invalid names is the intended behaviour.

lib/mobius/validations/event_validator.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Mobius.Validations.EventValidator do
+  @moduledoc false
+
+  alias Mobius.Core.Event
+  alias Mobius.Validations.Utils
+
+  @spec validate_names([any]) :: Utils.out()
```
Blond11516, line 7:
```suggestion
  @spec validate_names([Event.names()]) :: Utils.out()
```

We're expecting these to be event names aren't we?

test/mobius/actions/events_test.exs
```diff
@@ -0,0 +1,69 @@
...
+      assert {:errors, errors} = Events.subscribe([:THIS_ISNT_AN_EVENT])
+      assert ["invalid event name: :THIS_ISNT_AN_EVENT"] = errors
+    end
+
+    test "returns :ok if all event names are valid" do
+      assert :ok == Events.subscribe([:GUILD_BAN_ADD, :INVITE_CREATE, :USER_UPDATE])
+    end
+
+    test "sends subscribed event to caller as a message" do
+      Events.subscribe([:CHANNEL_CREATE, :WEBHOOKS_UPDATE, :CHANNEL_DELETE])
```
Blond11516: Why do you subscribe to 3 events?

test/mobius/core/event_test.exs
```diff
@@ -0,0 +1,50 @@
...
+
+    test "returns nil for event name atoms" do
+      assert nil == Event.parse_name(:CHANNEL_CREATE)
+    end
+
+    test "returns nil for non-event string" do
+      assert nil == Event.parse_name("THIS IS NOT AN EVENT")
+    end
+
+    test "returns nil for number" do
```
Blond11516: Are this and "returns nil for event name atoms" really necessary? While you're at it, why don't you do the same tests for tuples, maps, etc. ?

Same comment for the `is_event_name?/1` tests.

test/mobius/core/event_test.exs
```diff
@@ -0,0 +1,50 @@
...
+    test "returns false for non-atoms" do
+      assert not Event.is_event_name?(32)
+    end
+
+    test "returns false for invalid event name atom" do
+      assert not Event.is_event_name?(:THIS_IS_NOT_AN_EVENT)
+    end
+  end
+
+  describe "parse_data/2" do
```
Blond11516, line 46: What happens if we send an invalid name?

lib/mobius/core/event.ex
ZeLarpMaster: The events we receive from Discord are in all caps, keeping them as caps makes it a bit easier to deal with, though I guess we could lowercase it all 🤷

lib/mobius/core/event.ex
ZeLarpMaster: Yep, that's the intended behaviour. And indeed needs some docs. The idea is whoever is calling this function is expected to parse the event name first and do checks on that. If the event name is parsed and this function is called, it crashing here makes it easier to debug than the event going through while being possibly invalid.

lib/mobius/validations/event_validator.ex
ZeLarpMaster: We're expecting them to be, but it can be anything. If it's an event name, it spits out an `:ok`, otherwise it spits out an `{:errors, [String.t()]}` tuple. So the input here is any even if a "good" case is for `Event.names()` to be given

test/mobius/actions/events_test.exs
ZeLarpMaster: To show events from all of them are sent, at the end of the day it doesn't really matter much how many are here as long as one of them is used in the test.

test/mobius/core/event_test.exs
ZeLarpMaster: Well the event name atom case is relevant since parsed names are atoms. That test shows you can't parse the event name multiple times, it will only work with the initial strings.

For the number thing, it's just to make sure it doesn't *crash* with non-strings/atoms. We could add tests for all possible data structures, but then that's way too many tests for very little value.

lib/mobius/validations/event_validator.ex
ZeLarpMaster: I changed the typing of `Mobius.Actions.Events` to match this

test/mobius/core/event_test.exs
ZeLarpMaster: Should I combine both tests into a `returns nil for non-strings` and do 5-10 asserts for each data type instead?

lib/mobius/core/event.ex
ZeLarpMaster: Added docs about it

test/mobius/core/event_test.exs
Blond11516: Yeah if you really want such a test I think that would be a better approach.

Blond11516 (dismissed):

Blond11516 (approved):

PR #14, opened 2020-12-04
fix: Switching gateway encoding to json

ZeLarpMaster:
Also added `mix lint`


Blond11516 (approved):

Blond11516:
Haven't tested, I'm trusting you that it works 😛 
Do we want to add `mix lint` to the CI?

ZeLarpMaster:
> Do we want to add `mix lint` to the CI?

It already runs credo in the CI, the reason for the mix alias is just cause I keep forgetting the credo command :P

PR #17, opened 2020-12-23
feat: Manage ratelimits in the REST client

ZeLarpMaster:
The goal of this PR is to make the library follow the requirements detailed in [Discord's HTTP ratelimit documentation](https://discord.com/developers/docs/topics/rate-limits#rate-limits). The only things left to implement later in that documentation page will be detecting when a token becomes invalid and preventing permission issues.

TODO:
- [x] Write `@doc` in `Mobius.Services.RestRatelimiter`
- [x] Tests for `Mobius.Rest.Middleware.Ratelimit`
- [x] Tests for `Mobius.Services.RestRatelimiter`

### DIFF
--- a/lib/mobius/services/rest_ratelimiter.ex
+++ b/lib/mobius/services/rest_ratelimiter.ex
@@ -12,11 +12,18 @@ defmodule Mobius.Services.RestRatelimiter do
     GenServer.start_link(__MODULE__, opts, name: __MODULE__)
   end
 
-  @doc "Updates a bucket's remaining ratelimit and tracks the route's bucket for later usage"
-  @spec update_ratelimit(any, String.t(), integer, integer) :: :ok
-  def update_ratelimit(route, bucket, remaining, reset_after) do
+  @doc "Updates a route's remaining ratelimit and tracks the route's bucket"
+  @spec update_route_ratelimit(any, String.t(), integer, integer) :: :ok
+  def update_route_ratelimit(route, bucket, remaining, reset_after) do
     reset = System.monotonic_time(:millisecond) + reset_after
     GenServer.call(__MODULE__, {:update, route, bucket, remaining, reset})
+  end
+
+  @doc "Updates the global ratelimit when it's exceeded with how long to wait for the reset"
+  @spec update_global_ratelimit(integer) :: :ok
+  def update_global_ratelimit(reset_after) do
+    reset = System.monotonic_time(:millisecond) + reset_after
+    GenServer.call(__MODULE__, {:update_global, reset})
   end
 
   @doc "Checks if the route's bucket is known and waits for it to be available if ratelimited"
@@ -46,7 +53,7 @@ defmodule Mobius.Services.RestRatelimiter do
   end
 
   @impl GenServer
-  def handle_call({:update, "global", _, _, reset}, _from, state) do
+  def handle_call({:update_global, reset}, _from, state) do
     Process.send_after(self(), :reset_global, time_until(reset))
 
     {:reply, :ok, %{state | global_limit: reset}}
--- a/lib/mobius/services/rest_ratelimiter.ex
+++ b/lib/mobius/services/rest_ratelimiter.ex
@@ -1,0 +1,83 @@
+defmodule Mobius.Services.RestRatelimiter do
+  @moduledoc false
+
+  use GenServer
+
+  alias Mobius.Services.ETSShelf
+
+  @type state :: %{bucket_map: %{}, global_limit: nil | integer}
+
+  @spec start_link(keyword) :: GenServer.on_start()
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @spec update_ratelimit(any, String.t(), integer, integer) :: :ok
+  def update_ratelimit(route, bucket, remaining, reset_after) do
+    reset = System.monotonic_time(:millisecond) + reset_after
+    GenServer.call(__MODULE__, {:update, route, bucket, remaining, reset})
+  end
+
+  @spec wait_ratelimit(any) :: :ok
+  def wait_ratelimit(route) do
+    case GenServer.call(__MODULE__, {:request, route}) do
+      :ok -> :ok
+      {:wait, time} -> Process.sleep(time)
+    end
+  end
+
+  @impl GenServer
+  @spec init(keyword) :: {:ok, state()}
+  def init(_opts) do
+    state = %{
+      bucket_map: %{},
+      global_limit: nil
+    }
+
+    :ok = ETSShelf.create_table(__MODULE__, [:set, :protected])
+    {:ok, state}
+  end
+
+  @impl GenServer
+  def handle_info(:reset_global, state) do
+    {:noreply, %{state | global_limit: nil}}
+  end
+
+  @impl GenServer
+  def handle_call({:update, "global", _, _, reset}, _from, state) do
+    Process.send_after(self(), :reset_global, time_until(reset))
+
+    {:reply, :ok, %{state | global_limit: reset}}
+  end
+
+  def handle_call({:update, route, bucket, remaining, reset}, _from, state) do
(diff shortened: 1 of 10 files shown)


Blond11516 (commented):
Nothing to say for now! :+1:

test/mobius/services/rest_ratelimiter_test.exs
```diff
@@ -0,0 +1,53 @@
...
+  use ExUnit.Case
+
+  import Mobius.Assertions
+
+  alias Mobius.Services.RestRatelimiter
+
+  describe "wait_ratelimit/1" do
+    test "doesn't wait for global if never updated" do
+      assert_wait_time("global", 5)
+    end
```
Blond11516, line 11: So "doesn't wait" == "wait 5 ms"?

lib/mobius/services/rest_ratelimiter.ex
```diff
@@ -0,0 +1,85 @@
...
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @doc "Updates a bucket's remaining ratelimit and tracks the route's bucket for later usage"
+  @spec update_ratelimit(any, String.t(), integer, integer) :: :ok
+  def update_ratelimit(route, bucket, remaining, reset_after) do
+    reset = System.monotonic_time(:millisecond) + reset_after
+    GenServer.call(__MODULE__, {:update, route, bucket, remaining, reset})
+  end
```
Blond11516: What would you think of an "update_global_ratelimit" function so we can hide the awkward `update_ratelimit("global", "global", ...)` calls?

lib/mobius/services/rest_ratelimiter.ex
```diff
@@ -0,0 +1,85 @@
...
+    end
+  end
+
+  def handle_call({:request, route}, _from, state) do
+    now = System.monotonic_time(:millisecond)
+
+    case :ets.lookup(__MODULE__, find_bucket(state, route)) do
+      [] -> {:reply, :ok, state}
+      [{_, _, expiration}] when now > expiration -> {:reply, :ok, state}
+      [{_, remaining, _}] when remaining > 0 -> {:reply, :ok, state}
```
Blond11516, line 81: Shouldn't we check `expiration` and `remaining` together? Should we reply `:ok` if `now > expiration` but `remaining == 0`?

test/mobius/rest/middleware/ratelimit_test.exs
```diff
@@ -0,0 +1,98 @@
...
+    assert_get_time(ctx.client, url, 50)
+  end
+
+  test "waits until reset-after if ratelimit was exceeded", ctx do
+    url = "https://discord.com/api/v6/something"
+    mock(fn %{url: ^url} -> response(-1, 50, 0) end)
+    Tesla.get(ctx.client, url)
+
+    assert_get_time(ctx.client, url, 50)
+  end
```
Blond11516: Are those really different? Isn't the edge case `< 0` for remaining anyway?

test/mobius/rest/middleware/ratelimit_test.exs
```diff
@@ -0,0 +1,98 @@
...
+
+  defp assert_get_time(client, url, expected_time) do
+    client
+    |> Tesla.get(url)
+    |> function_time()
+    |> assert_in_delta(expected_time, 10)
+  end
+
+  defp response(remaining, delay, global_delay) do
```
Blond11516: I think this function would gain from taking a keyword list as a parameter instead of 3 distinct values. Right now there's no way to know what the tests are doing without looking at the implementation. Maybe a more descriptive name would help too.

Blond11516: More that that, it might be worth splitting it into multiple functions like `response_with_global_ratelimit` and `response_with_route_ratelimit` for example.

In many cases, especially when giving a global delay > 0, you don't actually care what the other values are (expect maybe for "asserting" that they are ignored), so why specify them at all?

test/support/assertions.ex
```diff
@@ -0,0 +1,21 @@
...
+  @spec function_time(any) :: Macro.output()
+  defmacro function_time(function) do
+    quote do
+      fn -> unquote(function) end
+      |> :timer.tc()
+      |> elem(0)
+      |> :erlang.convert_time_unit(:microsecond, :millisecond)
+    end
+  end
+end
```
Blond11516: That's not really an assertion...

test/mobius/services/rest_ratelimiter_test.exs
ZeLarpMaster: Need some tolerance for the execution time otherwise the test starts to fail randomly. `assert_wait_time` adds a `+- 10ms` to the expected time. So "doesn't wait" is actually "executes in less than 15ms" which is a fair assert imo for a function which would likely wait way more if something wrong were to happen.

lib/mobius/services/rest_ratelimiter.ex
ZeLarpMaster: I think that's a pretty good idea, I'll do that

lib/mobius/services/rest_ratelimiter.ex
ZeLarpMaster: You mean condense both together with an `or`? We could, but imo it's a bit more readable this way as it makes the lines less dense.

test/mobius/rest/middleware/ratelimit_test.exs
ZeLarpMaster: Oh you're right, I derped while copy pasting stuff :facepalm:

test/mobius/rest/middleware/ratelimit_test.exs
ZeLarpMaster: That's a good idea, it didn't occur to me how unreadable those tests were after spending so much time on them

test/support/assertions.ex
ZeLarpMaster: That's very true :joy: It was at first, but isn't anymore so I'll move it elsewhere

test/mobius/services/rest_ratelimiter_test.exs
ZeLarpMaster: > otherwise the test *could start to fail randomly

This is more likely to happen on the CI since its processing power might be fairly limited. Maybe the tolerance is too high? I'm not sure what I'd change here to make it clearer or better :thinking: Do you have suggestions?

Blond11516: Not really, timing based tests are always kind of a pain so that seems good enough.

lib/mobius/services/rest_ratelimiter.ex
Blond11516: My original comment wasn't that clear but I'm mostly wondering if the current behaviour is correct or not.

The way the code is currently written, we could let a call through because `now > expiration` even though `remaining <= 0`. Am I correct? If so, is this even possible or do Discord's ratelimiting rules prevent from happening in the first place?

ZeLarpMaster: Well when `now > expiration`, the latest `update_route_ratelimit` has expired. This means the bucket's ratelimit has reset and is ready to accept new requests again. So even if `remaining <= 0`, the bucket is actually "full" again at that point in time.

For example, consider the following sequence of events:
- User makes a request
- Discord replies with the HTTP headers `x-ratelimit-remaining: 0` and `x-ratelimit-reset-after: 500`
- Nothing happens for more than 500ms
- User makes another request on the same bucket

At this point in time, the bucket has reset as specified by the `reset-after`. This means the `remaining` resets back to the limit of `5` and the new request will go through, replying with the header `x-ratelimit-remaining: 4` and some reset time in the future.

I hope this makes sense!

Blond11516: Yup, makes perfect sense! I just wasn't sure what the exact ratelimiting rules were.

Blond11516 (approved):

PR #18, opened 2020-12-29
feat: Added shard commands ratelimit

ZeLarpMaster:
[This ratelimit](https://discord.com/developers/docs/topics/gateway#rate-limiting) has a little quirk even though it isn't documented: it doesn't simply reset after 60 seconds (the way the ratelimit works in #17). Instead, it acts like the [sliding window log described in this article](https://medium.com/figma-design/an-alternative-approach-to-rate-limiting-f8a06cf7c94c).

The difference between a timed reset and this "self-refill" (as it's called in the code) is a timed reset will allow more than the limit along its border. For example, with the limit of 120 commands every 60 seconds on a timed reset approach, a user could do the following:
- Do one request
- Wait 50 seconds
- Do a burst of 110 commands within the next 10 seconds
- Wait a few seconds
- Do a burst of 50 commands within the next 10 seconds
During the last burst of commands, Discord will disconnect the bot from the gateway because it will consider the ratelimit as being exceeded. This means Discord never wants more than 120 commands within the last 60 seconds of any point in time. This fits the behavior of the sliding window log. The "self-refill" approach mimics this behavior using BEAM's message passing and token counters.

Alternatively, it can be ratelimited by allowing only one command every 500 milliseconds, but this approach doesn't let the user do bursts of requests when needed which would be an artificial limit set by the library. The design decision was therefore to allow as much freedom as possible.

TODO:
- [x] Add tests for `Mobius.Services.ShardRatelimiter.SelfRefill`
- [x] Add tests for `Mobius.Actions.Status` about the ratelimit
- [x] Find a better name for `ShardRatelimiter`
- [x] Try it and make sure it works properly

### DIFF
--- a/lib/mobius/actions/status.ex
+++ b/lib/mobius/actions/status.ex
@@ -6,7 +6,7 @@ defmodule Mobius.Actions.Status do
   alias Mobius.Core.BotStatus
   alias Mobius.Core.Opcode
   alias Mobius.Services.Bot
-  alias Mobius.Services.ShardRatelimiter
+  alias Mobius.Services.CommandsRatelimiter
   alias Mobius.Services.Socket
 
   @doc """
@@ -21,8 +21,8 @@ defmodule Mobius.Actions.Status do
       # The "update_status" ratelimit of 5 per minute was found by testing a lot
       # because this limit is undocumented in the official documentation of the Discord API
       # This limit was also "confirmed" by another user on the unofficial Discord API server
-      with :ok <- ShardRatelimiter.request_access(shard, "update_status", 60_000, 5),
-           :ok <- ShardRatelimiter.request_access(shard) do
+      with :ok <- CommandsRatelimiter.request_access(shard, "update_status", 60_000, 5),
+           :ok <- CommandsRatelimiter.request_access(shard) do
         new_status
         |> Opcode.update_status()
         |> Socket.send_message(shard)
--- a/lib/mobius/actions/status.ex
+++ b/lib/mobius/actions/status.ex
@@ -6,6 +6,7 @@ defmodule Mobius.Actions.Status do
   alias Mobius.Core.BotStatus
   alias Mobius.Core.Opcode
   alias Mobius.Services.Bot
+  alias Mobius.Services.ShardRatelimiter
   alias Mobius.Services.Socket
 
   @doc """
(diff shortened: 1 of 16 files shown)


lib/mobius/actions/status.ex
```diff
@@ -14,12 +15,20 @@ defmodule Mobius.Actions.Status do
...
-  @spec change_status(BotStatus.t()) :: [:ok]
+  @spec change_status(BotStatus.t()) :: [:ok | {:error, any}]
   def change_status(%BotStatus{} = new_status) do
     for shard <- Bot.list_shards() do
-      new_status
-      |> Opcode.update_status()
-      |> Socket.send_message(shard)
+      # The "update_status" ratelimit of 5 per minute was found by testing a lot
+      # because this limit is undocumented in the official documentation of the Discord API
+      # This limit was also "confirmed" by another user on the unofficial Discord API server
```
Blond11516, line 23: Is the Discord staff aware of this? Is there an official channel to notify them of this lack of documentation?

Blond11516 (commented):
No comment so far :+1:

lib/mobius/actions/status.ex
ZeLarpMaster: https://github.com/discord/discord-api-docs/issues is the way to go. Not sure they'd care, but it might be worth a try :shrug:

Blond11516: I'll let you see if you want to report it :stuck_out_tongue:

lib/mobius/core/shard_buckets.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Mobius.Core.ShardBuckets do
+  @moduledoc false
+
+  @type t :: %{optional(String.t()) => non_neg_integer}
```
Blond11516, line 4: I've never seen `optional` before and not quite sure that I understand what it does. The docs say "map with optional pairs of key_type and value_type" (vs. "map with required pairs of key_type and value_type" for a typespec without `optional`), but couldn't you have an empty map even without optional? Is that not what it's for?

Blond11516 (approved):

lib/mobius/core/shard_buckets.ex
ZeLarpMaster: For generic types like `String.t()` or `atom`, I honestly have no idea whether there's a difference between `optional` and `required`. In both cases it seems to mean "all keys are of type X". But for literal types it means what you'd expect: `%{optional(:key) => integer}` means the map can be empty or have the key `:key` with a value of type integer.

So in my mind I still assume maps with `required` to never be empty and maps with `optional` to possibly be empty, but I don't think the typespec enforces that.

PR #19, opened 2020-12-30
feat: Cogs

Blond11516:
TODO:
- [x] Write documentation for the Cog module
- [x] Add support for simple command handlers
- [x] Add a cog loader to automatically load static cogs
- [x] Add tests


lib/mobius/services/cog.ex
```diff
@@ -0,0 +1,89 @@
+defmodule Mobius.Services.Cog do
+  defmacro __using__(_call) do
+    quote do
+      use GenServer
+      require Logger
+
+      alias Mobius.Actions.Events
+
+      :persistent_term.put(:event_handlers, %{})
```
ZeLarpMaster: `:persistent_term` is global to all processes, so a key like `:event_handlers` without a namespace isn't exactly a best practice (see [the erlang docs](https://erlang.org/doc/man/persistent_term.html#best-practices-for-using-persistent-terms))

While you could namespace it, I don't really like the idea of using global state like this especially since anyone could shoot themselves in the foot by modifying it. I think it would be better to use [module attributes](https://hexdocs.pm/elixir/Module.html#register_attribute/3) instead to store the event handlers during compilation.

lib/mobius/cogs/test_cog.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Mobius.Cogs.TestCog do
+  use Mobius.Services.Cog
```
ZeLarpMaster: The module itself isn't exactly a service, but it also doesn't feel like core, or actions. So I feel like `Mobius.Cog` might be more appropriate?

lib/mobius/services/cog.ex
```diff
@@ -0,0 +1,89 @@
...
+
+      import unquote(__MODULE__), only: [listen: 2, listen: 3]
+
+      # TODO: replace &IO.inspect/1 with command handler
+      listen :message_create, %{"content" => content} do
```
ZeLarpMaster: That's such a good idea. I thought I'd split commands into a "command parsing service", but that's actually a much better idea since it puts the work on the cog process instead and doesn't involve a bunch of message passing to grab information from here and there. The only downside is it potentially duplicates some work, but honestly I don't think there's that much work happening in multiple places so it should be fine.

lib/mobius/services/cog.ex
```diff
@@ -0,0 +1,89 @@
...
+      def init(_opts) do
+        event_names =
+          :persistent_term.get(:event_handlers)
+          |> Enum.map(&elem(&1, 0))
+
+        Events.subscribe(event_names)
+
+        Logger.debug("Cog \"#{__MODULE__}\" subscribed to events #{inspect(event_names)}")
+
+        {:ok, nil}
```
ZeLarpMaster: I wonder if we'll ever want an API to support the state here instead of always through ets tables :thinking: Definitely not for now or here, but it's a thought

lib/mobius/services/cog.ex
```diff
@@ -0,0 +1,89 @@
...
+      end
+
+      @impl true
+      def handle_info({event_name, data}, state) do
```
ZeLarpMaster: Makes me wonder if we'd want a special key prefix to put in these messages so they don't collide with other messages. I realize a 2 element tuple is kinda generic and could easily happen with things like `{:ok, any}`.

lib/mobius/services/cog.ex
```diff
@@ -0,0 +1,89 @@
...
+          quote do
+            unquote(block)
+            :ok
+          end
+
+        _ ->
+          quote do
+            try(unquote(contents))
+            :ok
+          end
```
ZeLarpMaster: What does all this transformation of `contents` do? It looks to me like it only makes it always return `:ok`, but I'm not sure of why that's necessary

lib/mobius/services/cog.ex
ZeLarpMaster: Looking back at how it's used, I think you can replace this one with `Module.register_attribute/3`, the `:persistent_term.get` with `Module.get_attribute/3` and the persistent term update with `Module.put_attribute/3` and I feel like this should work :thinking:

lib/mobius/services/cog.ex
```diff
@@ -0,0 +1,89 @@
...
+
+      name = :"#{Atom.to_string(event_name)}#{Integer.to_string(handler_id)}"
+
+      updated_handlers =
+        :persistent_term.get(:event_handlers)
+        |> Map.update(event_name, [name], fn handlers ->
+          [name | handlers]
+        end)
+
+      :persistent_term.put(:event_handlers, updated_handlers)
```
ZeLarpMaster: Doesn't this quote do too much? All of that code ends up executed in the module afterwards, but I feel like this could run inside the macro instead :thinking: I know it isn't leaking anything into the module, but it still feels weird.

ZeLarpMaster (commented):
Overall that's a pretty good start and hypes me up because this means we're getting close to a v1 :tada:

lib/mobius/services/cog.ex
Blond11516: I had tried to do it with module attributes at first but couldn't get it to work (ping me if you want to know the details). I've got it working now with `@before_compile`!

lib/mobius/cogs/test_cog.ex
Blond11516: Yeah I wasn't sure where to put it. `Mobius.Cog` feels good to me.

lib/mobius/services/cog.ex
Blond11516: Yeah the gist of command management can still be written in a dedicated module, but I think having the entrypoint here is going to be the simplest solution by far.

lib/mobius/services/cog.ex
Blond11516: That's a good point. Another solution would be to generate a bunch of different `handle_info` definitions to pattern match message separately. I don't think that would be too hard to do :thinking:

lib/mobius/services/cog.ex
Blond11516: Yeah I thought about that too. I feel like it would be a nice option to have, but off the top of my head I fail to see a use case where you'd want a state but wouldn't rather have it be persisted in an ETS...

lib/mobius/services/cog.ex
Blond11516: TBH I'm not too sure, but `contents =` through `contents = Macro.escape(contents, unquote: true)` is copy-pasted from ExUnit, so I assume it's necessary :sweat_smile: I'll do some testing to try and figure out exactly what this whole portion does.

lib/mobius/services/cog.ex
Blond11516: I agree it's kind of heavy, but all all attribute manipulation has to happen inside the returned AST, so anything that manipulates the `event_handlers` has to be quoted.

lib/mobius/services/cog.ex
ZeLarpMaster: There's definitely cases where we don't want persistance (like temporary caching or a state for tasks), but that'll come later

lib/mobius/services/cog.ex
ZeLarpMaster: I think a special key prefix (ie.: `{:discord_event, event_name, data}`) might be simpler and easier to read than some macro trickery to generate callbacks + it avoids unexpected behavior where a cog `send/2` itself a tuple, but doesn't receive the answer in its `handle_info`

That's also the way [GenServer distinguishes its `call` and `cast` from the other messages](https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen_server.erl#L875)

lib/mobius/services/cog.ex
ZeLarpMaster: It might be how `test` supports the [implicit try](https://hexdocs.pm/elixir/Kernel.html#def/2-rescue-catch-after-else). My hypothesis on why they did all of that is they might've needed it just to add the `:ok`, but since we don't need any of it, we might be fine passing `contents` along directly as `def/2`'s 2nd argument (ie.: `def unquote(event_name)(unquote(var)), unquote(contents)`). This might be worth the try. If it doesn't work, I'd suggest either extracting this into a function with comments explaining its purpose or simply adding comments about the why.

Blond11516: You were right, I just removed the whole `case` and everything seems to be working fine.

.formatter.exs
```diff
@@ -1,4 +1,14 @@
 # Used by "mix format"
+locals_without_parens = [
+  # Mobius.Services.Cog
```
ZeLarpMaster: `Mobius.Cog` now

lib/mobius/cog.ex
```diff
@@ -0,0 +1,150 @@
...
+
+  defmacro __before_compile__(_env) do
+    quote do
+      use GenServer
+      alias Mobius.Actions.Events
+
+      listen :message_create, %{"content" => content} do
+        case Command.parse_command(@commands, content) do
+          {:ok, command, arg_values} ->
+            Command.execute(command, arg_values)
```
ZeLarpMaster: I think this whole thing would be simpler if a single function in Command was doing all of that (`Command.parse_command/2`, `Command.execute/2`, and the argument error handling). The function could return `{:ok, return_value_of_the_command} | {:error, reason}`. It would reduce the amount of logic required in `Cog` which in my opinion is a win.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+          {:too_few_args, command, received} ->
+            Logger.info(
+              "Too few arguments for command \"#{command.name}\". Expected #{
+                Command.arg_count(command)
+              } arguments, got #{received}."
+            )
+
+          {:invalid_args, errors} ->
+            Enum.each(errors, fn {{arg_name, arg_type}, value} ->
+              # credo:disable-for-next-line Credo.Check.Readability.StringSigils
```
ZeLarpMaster: I think it would be better to fix this instead of disabling it

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
+defmodule Mobius.Cog do
+  @moduledoc """
+  Defines a cog.
+
+  Cogs are small, self-contained services that implement a bot's logic. Cogs
+  have two ways of interacting with Discord: `listen/3` and `command/3`.
```
ZeLarpMaster: > Cogs have two ways of interacting with Discord

I see what you mean, but that's not technically right which might lead to confusion. Those are the two ways for the cog to bind itself to events. Cogs can interact with Discord through everything in `Mobius.Actions` which will eventually be more than 100 functions.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+  `listen/3` instructs the cog to listen to specific events that are sent from
+  Discord's API. These events are generally the result of normal user
+  interaction in Discord.
```
ZeLarpMaster: > These events are generally the result of normal user interaction in Discord.

While that's true, I think we can be a bit more specific. Maybe something along the lines of

> Events are sent when users (humans or bots) interact with Discord. These include almost everything you can do on Discord from messages sent to users changing their username. For a complete list of events and the data associated with them see the Events documentation page.

The Events documentation page doesn't exist yet, but I think it would be the best place to put all of that information instead of having it in a random module somewhere.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+  `command/3` defines a specific command that users can enter in the server's
+  textual chat to manually trigger an interaction with the bot.
```
ZeLarpMaster: Bots aren't necessarily in only one server so it would be more accurate to say something like the following. I also think it might be worth being a little bit more specific about the interaction with the bot.

> `command/3` defines a command which can be used by users in text channels where the bot has the permission to read. These commands will execute the code specified in the `do` block of the command with its arguments.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+  ## Example
+  ```elixir
+  defmodule MyCog do
+    use Mobius.Cog
+
+    listen :guild_member_add, %{"user" => user} do
+      Logger.puts("Welcome \#{user["username"]}")
```
ZeLarpMaster: I don't think `Logger.puts/1` exists. I assume you changed it from `IO.puts` :P

Also the `\` before the `#` probably makes this line not compile. In general I think we need to be careful with code in user-visible docs because people will definitely copy paste stuff from the docs and it's likely to cause trouble if the code doesn't run.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+      |> String.trim()
+      |> IO.puts()
+    end
+  end
+  ```
+
+  This cog does the following things:
+  * Every time a new user joins the server, it greets the user
+  * Every time a user enters "repeat word times", it repeats the word "word"
+    "times" times
```
ZeLarpMaster: It would probably be even better if these explanations were directly inside the code snippet as comments. This way if people copy paste it, they'll grab the comments with it which would help them better understand exactly which parts do what.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+        {:noreply, state}
+      end
+    end
+  end
+
+  @doc """
+  Defines an event handler for a Discord event.
+
+  The second parameter, "payload", can be used to access the events payload.
```
ZeLarpMaster, line 136: Just like above, we could "link" to the Events page which includes all possible events even though it doesn't exist yet.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+    IO.puts("Message received: \#{content}")
+  end
+  ```
+  """
+  defmacro listen(event_name, payload \\ quote(do: _), do: block) do
+    payload = Macro.escape(payload)
+    contents = Macro.escape(block, unquote: true)
+
+    quote bind_quoted: [event_name: event_name, payload: payload, contents: contents] do
+      existing_handlers = Module.get_attribute(__MODULE__, :event_handlers)
```
ZeLarpMaster, line 152: Out of curiosity, can you use the `@` notation here or does it not work? I feel like `@event_handlers` would be easier to read here than all of this, but I'm not sure whether that's actually possible or not 🤔

ZeLarpMaster: The `put_attribute` would become `@event_handlers {event_name, Function.capture(__MODULE__, name, 1)}`

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+      def unquote(name)(unquote(payload)), do: unquote(contents)
+    end
+  end
+
+  @doc """
+  Defines a command to be used by Discord users.
+
+  The first parameter defines the name of the command as a single word (no
+  spaces).
```
ZeLarpMaster, line 170: You don't have to do it in this PR, but it would be very nice if we could `raise` an exception in the macro if this requirement isn't met such that it fails at compile time which makes it very easy to debug.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+  ## Command types
+
+  The list of supported argument types is as follows:
+  * `:string`
+  * `:integer`
+
+  If you wish to use a type that isn't supported by Mobius, `:string` should be
+  used as a "catch all" type. You can then implement your own validation as part
+  oh the command body.
```
ZeLarpMaster: typo
> as part __oh__ the command body

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+  If a user enters a command with an argument that doesn't match its
+  corresponding type, Mobius will automatically reply to let the user know what
+  the expected type is.
+
+  ## Duplicate commands
+
+  Should you define two commands with the same name, Mobius will raise a
+  compilation warning indicating that the second command will be ignored.
```
ZeLarpMaster: I think this should be a compilation error. Two commands with the same name on different cogs could be a warning (to be implemented later), but two in the same cog is likely a programmer error (copy pasting stuff and forgetting to change the name of the command) and treating it as such makes it easier to debug than a warning which might be missed during compilation.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+  ## Example
+  ```elixir
+  command "add", num1: :integer, num2: integer do
```
ZeLarpMaster: typo, missing a `:` before the 2nd `integer` type

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+    num1 + num2
+  end
+  ```
+
+  In discord chat:
+  ```
+  user: add 1 2
+  myBot: 3
+  user: add 1 hello
+  myBot: "Invalid type for argument "num2". Expected "integer", got "hello".
```
ZeLarpMaster: While it might not be the term used colloquially, I think it's better to use technical terms like `In a Discord text channel` so people reading this can connect the dots between the concepts easily.

Also I think it's a typo, but the `"` before `Invalid type` should probably be removed (since it isn't closed at the end and wouldn't be in the message sent)

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+  """
+  defmacro command(command_name, args \\ [], do: block) do
+    handler_name = Command.command_handler_name(command_name)
+
+    new_command = %Command{
+      name: command_name,
+      args: args,
+      handler: Function.capture(__CALLER__.module, handler_name, length(args))
```
ZeLarpMaster, line 211: I wonder if it would change anything if the `handler` was an anonymous function instead of a named and `def`'d function. It would definitely prevent all possible function name collisions, but maybe it silences compilation errors and stuff like that? This is just a thought though, probably not worth changing this.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+            arg_vars: arg_vars,
+            command: Macro.escape(new_command)
+          ] do
+      existing_commands = Module.get_attribute(__MODULE__, :commands)
+
+      if Module.defines?(__MODULE__, {handler_name, 0}) do
+        IO.warn(
+          "Command \"#{command.name}\" already exists. Duplicated command will be ignored.",
+          Macro.Env.stacktrace(__ENV__)
+        )
```
ZeLarpMaster, line 241: Is the warning [in the docs](https://hexdocs.pm/elixir/IO.html#warn/1) relevant here? Since it's a macro I think maybe not, but I'm really not sure.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+    end
+  end
+
+  @doc false
+  def event_handler_name(event_name, event_handlers) do
+    handler_id =
+      event_handlers
+      |> Enum.group_by(&elem(&1, 0))
+      |> Map.get(event_name, [])
+      |> length()
```
ZeLarpMaster: Would `Enum.count(event_handles, & elem(&1, 0) == event_name)` work here?

lib/mobius/cog.ex
```diff
@@ -0,0 +1,243 @@
...
+
+    :"mobius_event_#{Atom.to_string(event_name)}_#{Integer.to_string(handler_id)}"
```
ZeLarpMaster: I don't think you need to convert the types before interpolating them:
```elixir
iex> :"something_#{:hello}_#{1}"
:something_hello_1
```

lib/mobius/cogs/ping_pong.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Mobius.Cogs.PingPong do
+  @moduledoc false
+
+  use Mobius.Cog
+
+  require Logger
+
+  command "ping", do: Logger.info("pong")
```
ZeLarpMaster, line 8: I love this cog 😂

lib/mobius/core/command.ex
```diff
@@ -0,0 +1,99 @@
+defmodule Mobius.Core.Command do
+  @moduledoc false
+
+  alias Mobius.Command.ArgumentParser
```
ZeLarpMaster: The module should probably be at `Mobius.Core.Command.ArgumentParser` or even `Mobius.Core.ArgumentParser` instead

lib/mobius/core/command.ex
```diff
@@ -0,0 +1,99 @@
...
+          args: keyword(ArgumentParser.arg_type()),
+          handler: function()
+        }
+
+  @spec command_handler_name(String.t()) :: atom()
+  def command_handler_name(command_name) do
+    :"mobius_command_#{command_name}"
+  end
+
+  @spec get_command_arg_names(keyword(atom())) :: [atom()]
```
ZeLarpMaster: Isn't the type here `keyword(ArgumentParser.arg_type())` instead of `keyword(atom())`? I know it's a long type, but you could probably define it at the top as `@type args :: keyword(ArgumentParser.arg_type())`

lib/mobius/core/command.ex
```diff
@@ -0,0 +1,99 @@
...
+  def get_command_arg_names(args) do
+    Enum.map(args, &elem(&1, 0))
+  end
+
+  @spec parse_command([t()], String.t()) ::
+          :not_a_command
+          | {:ok, t(), [String.t()]}
+          | {:too_few_args, t(), non_neg_integer()}
+          | {:invalid_args, [{Validator.arg_type(), String.t()}]}
```
ZeLarpMaster: Since this return type could possibly be useful elsewhere (and also because it's very long), I think it would be better if it was defined at the top so it could be used in other modules without copy pasting all of this.

lib/mobius/core/command.ex
```diff
@@ -0,0 +1,99 @@
...
+  def parse_command(commands, message) do
+    case Enum.find(commands, fn %__MODULE__{} = command ->
+           String.starts_with?(message, command.name)
+         end) do
```
ZeLarpMaster: Would probably be slightly more readable by splitting this line with pipes:
```elixir
commands
|> Enum.find(fn ... end)
|> case do
```

lib/mobius/core/command.ex
```diff
@@ -0,0 +1,99 @@
...
+      nil ->
+        :not_a_command
+
+      %__MODULE__{} = command ->
+        arg_values =
+          message
+          |> String.split()
```
ZeLarpMaster: It's not the case yet, but the parsing logic will eventually be smarter than just splitting on spaces, so it would be nice (not necessary) if this could be either inside `validate` or going through another function dedicated to parsing the argument strings.

lib/mobius/core/command.ex
```diff
@@ -0,0 +1,99 @@
...
+          |> tl()
+
+        with {:ok, values} <- validate(command, arg_values) do
+          {:ok, command, values}
+        end
+    end
+  end
+
+  @spec validate(t(), [String.t()]) ::
+          {:ok, [any()]} | {:too_few_args, t(), non_neg_integer()} | {:invalid_args, [String.t()]}
```
ZeLarpMaster: I don't remember why or what exactly (was it the compiler or ElixirLS?), but I know something was complaining about specs on private functions. I should look into it, but you can pretty much remove the spec here.

lib/mobius/core/command/argument_parser/integer.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Mobius.Command.ArgumentParser.Integer do
+  @moduledoc false
+
+  @behaviour Mobius.Command.ArgumentParser
+
+  @impl true
+  def parse(value) do
+    case Integer.parse(value) do
+      :error -> :error
+      {int, _} -> int
```
ZeLarpMaster: This will return `5` for the value `"5.0"` which is not exactly what we want. It can be fixed with:

```suggestion
      {int, ""} -> int
      _ -> :error
```

lib/mobius/services/cog_loader.ex
```diff
@@ -0,0 +1,34 @@
...
+  use GenServer
+
+  @initial_cogs [Mobius.Cogs.PingPong]
+
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @spec load_cog(module()) :: DynamicSupervisor.on_start_child()
```
ZeLarpMaster: I'm assuming it was forgotten while refactoring, but `load_cog/1` actually returns `:ok`

lib/mobius/services/cog_loader.ex
```diff
@@ -0,0 +1,34 @@
...
+  def load_cog(cog) do
+    GenServer.call(__MODULE__, {:load_cog, cog})
+  end
+
+  @impl true
+  def init(_init_arg) do
+    Enum.each(@initial_cogs, &start_cog/1)
+
+    {:ok, @initial_cogs}
```
ZeLarpMaster: Do you think this service could ever have more than the list of cogs in its state? I feel like we should future-proof it and wrap the list in a simple map 🤔

ZeLarpMaster (changes requested):
Whew! That's a lot of comments! Overall I really like where this is going, the architecture looks very good though I foresee a couple of changes in here in a decently far future, but it shouldn't be too bad. Obviously a lot of these comments are very nitpicky and this is a very important chunk of code, so I hope you won't feel too bad about this :P

lib/mobius/cog.ex
Blond11516: Right, at first I thought you couldn't use interpolation in the string sigil but I had forgotten about the lowercase version. Will fix!

lib/mobius/cog.ex
Blond11516: Another option for the Event documentation (and all kinds of general documentation) would be a "Guides" section in the docs, like Phoenix does for example.

I think this would be an interesting option because I don't think we intend for the `Events` module to be "public", right?

lib/mobius/cog.ex
Blond11516: Yup, I wanted to write `IO` :stuck_out_tongue: 

I just generated the docs and `\#{user["username"]}` does get escaped properly to `#{user["username"]}`, but wrapping the doc in `~S` and getting rid of the `\` looks nicer anyway.

lib/mobius/cog.ex
Blond11516: I get a warning saying the attribute is undefined and then an error because something is `nil`. I never saw the `@attribute` syntax in any macro example I looked at, so I'm pretty sure it's just not allowed.

lib/mobius/cog.ex
Blond11516: Good point. I'll add a todo.

lib/mobius/cog.ex
Blond11516: Is it even worth documenting this then?

lib/mobius/cog.ex
Blond11516: Hmm that's an interesting point. I had originally went with named functions since I was using `apply/3` to call them (which uses the module and function name rather than the function itself), but it would indeed make the code simpler if we could use anonymous functions (since we could remove the counts and stuff). I'll look into this, I think it's worth it.

lib/mobius/cog.ex
Blond11516: The warning is for `warn/1`, not `warn/2`.

Also per your previous comment I changed it to an error so it doesn't matter either way :stuck_out_tongue:

lib/mobius/cog.ex
Blond11516: Looks like it works!

lib/mobius/cog.ex
ZeLarpMaster: Everything outside `Mobius.Actions` (and `Mobius.Cog` now I guess :P) won't be public indeed. I also wanted the Events page to be in a "Guides" section so we were on the same page about that!

lib/mobius/cog.ex
ZeLarpMaster: Good to know! Thanks

lib/mobius/cogs/ping_pong.ex
Blond11516: Haha thanks! :stuck_out_tongue: 

I removed my `TestCog` but I thought we could keep this as a sanity check, kind of like people often do with APIs.

lib/mobius/cog.ex
ZeLarpMaster: Probably not, you're right

lib/mobius/core/command.ex
Blond11516: Good point. Which do you prefer? The module names are getting kind of long for the implementation modules, but I feel like it really makes more sense under `Command` :thinking:

lib/mobius/core/command.ex
Blond11516: I don't really mind the long spec, it's not like short specs aren't common anyway :stuck_out_tongue:

lib/mobius/core/command.ex
ZeLarpMaster: I agree the names get long, but it usually only makes the alias longer so it shouldn't matter much. I prefer keeping Core flat if possible, but I'm also starting to think `Mobius.Core.Command` will likely have more modules as features are added so maybe it's better to keep them together 🤔 So keeping it under Command sounds better

lib/mobius/core/command.ex
Blond11516: I put it in a separate function. More functions = more better ;)

lib/mobius/core/command.ex
Blond11516: Would be nice if you could find what it was, I don't remember ever seeing an issue with that :thinking:

lib/mobius/core/command/argument_parser/integer.ex
Blond11516: Good catch!

lib/mobius/services/cog_loader.ex
Blond11516: This is indeed a refactoring error, but I wasn't sure what we wanted it to return. Is `:ok` good? Could we gain something from actually returning `on_start_child()`?

lib/mobius/services/cog_loader.ex
Blond11516: Sure, don't see why not

lib/mobius/cog.ex
Blond11516: I updated this part. Let me know what you think.

lib/mobius/services/cog_loader.ex
ZeLarpMaster: I think `:ok` is fine, the pid shouldn't ever be used directly since it might change if the cog restarts so that leaves us with the `:ok` or an error. So `:ok | {:error, any}` is what we want I think

lib/mobius/core/command.ex
ZeLarpMaster: I looked at ElixirLS, dialyzer, credo, and compilation. None of them seem to say anything about it so I think I was confusing it with `@doc`:
image

I consider `@spec`s to be documentation for the user of the function, but for private functions, we can't document the functions, so I probably never put any specs because of that. So far none of the private functions have a spec, so I don't know if we want to start writing specs everywhere starting now or continue not writing specs. I feel like you need to read private functions to know what they do anyway, so the specs don't add much value at the cost of bloating code a little bit (especially for functions which are one-liners) and I'd much prefer a hard, always followed, rule than an arbitrary "if function is big enough, put a spec"

This is up for debate and this PR is probably not the right place for such a debate. You can do whichever you want and we can talk about it on Discord

lib/mobius/cog.ex
ZeLarpMaster: Looks good!

lib/mobius/cog.ex
Blond11516: I've moved what I could to the `Command` module, but there really wasn't a lot of logic in that case statement anyway. Let me know what you think.

lib/mobius/core/command.ex
Blond11516: I removed the specs for private functions, I really don't have any strong feelings about this.

lib/mobius/cog.ex
Blond11516: Well, I can't get it to work, nor can I find any examples of this, so I'm gonna drop this for now.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,244 @@
...
+  ```elixir
+  defmodule MyCog do
+    use Mobius.Cog
+
+    # Every time a new user joins the server, greet the user
+    listen :guild_member_add, %{"user" => user} do
+      IO.puts("Welcome #{user["username"]}")
+    end
+
+    # Every time a user enters "repeat word times", repeatthe word "word"
```
ZeLarpMaster: Typo: missing a space between "repeat" and "the"

lib/mobius/services/cog_loader.ex
```diff
@@ -0,0 +1,38 @@
...
+    Enum.each(@initial_cogs, &start_cog/1)
+
+    {:ok, %{cogs: @initial_cogs}}
+  end
+
+  @impl true
+  def handle_call({:load_cog, cog}, _from, state) do
+    case start_cog(cog) do
+      {:ok, _pid} -> {:reply, :ok, %{state | cogs: [cog | state.cogs]}}
+      {:ok, _pid, _info} -> {:reply, :ok, %{state | cogs: [cog | state.cogs]}}
```
ZeLarpMaster: We're actually in control of what `Cog` returns on `init/2` so we know it's impossible for cogs to reply with `{:ok, _pid, _info}`. I'm not sure if this was added because of linters, but unless that's the case, I think you can remove this line.

ZeLarpMaster (commented):
Great work! Some stuff will eventually be moved around to support other features, but for now this looks very good!

lib/mobius/cog.ex
```diff
@@ -0,0 +1,244 @@
...
+      @impl true
+      def handle_info({event_name, data}, state) do
+        Logger.debug("Cog \"#{__MODULE__}\" received event #{inspect(event_name)}")
+
+        @event_handlers
+        |> Enum.filter(&match?({^event_name, _handler}, &1))
+        |> Enum.each(fn {_event_name, handler} -> apply(handler, [data]) end)
+
+        {:noreply, state}
+      end
```
ZeLarpMaster, line 129: Out of curiosity, what happens if the cog defines its own `handle_info`? Does it use the cog's before the generated one or the other way around? I think that might be worth documenting at the very least.

lib/mobius/cog.ex
```diff
@@ -0,0 +1,244 @@
...
+    end
+  end
+
+  @doc """
+  Defines a command to be used by Discord users.
+
+  The first parameter defines the name of the command as a single word (no
+  spaces).
+
+  The second parameter defires the list of arguments that the command accepts,
```
ZeLarpMaster: Typo: `defires` --> `defines`

lib/mobius/cog.ex
```diff
@@ -0,0 +1,244 @@
...
+          ] do
+      existing_commands = Module.get_attribute(__MODULE__, :commands)
+
+      if Module.defines?(__MODULE__, {handler_name, 0}) do
+        reraise(
+          %CompileError{
+            line: line,
+            file: file,
+            description:
+              "Command \"#{command.name}\" already exists. Duplicated command will be ignored."
```
ZeLarpMaster: `Duplicated command will be ignored.` Well technically it makes a compilation error so it isn't exactly ignored :P

lib/mobius/core/command.ex
```diff
@@ -0,0 +1,103 @@
...
+          name: String.t(),
+          args: keyword(ArgumentParser.arg_type()),
+          handler: function()
+        }
+
+  @type handle_message_result ::
+          {:ok, any()}
+          | :not_a_command
+          | {:too_few_args, t(), non_neg_integer()}
+          | {:invalid_args, [{{:atom, ArgumentParser.arg_type()}, String.t()}]}
```
ZeLarpMaster: `:atom` means the literal atom `:atom`, I think this is a typo for `atom()` 🤔

lib/mobius/core/command.ex
```diff
@@ -0,0 +1,103 @@
...
+    commands
+    |> Enum.find(fn %__MODULE__{} = command -> String.starts_with?(message, command.name) end)
+    |> case do
+      nil ->
+        :not_a_command
+
+      %__MODULE__{} = command ->
+        arg_values = split_arguments(message)
+
+        with {:ok, values} <- validate(command, arg_values) do
```
ZeLarpMaster: I think we could merge `validate` and `execute` here and it would make this module a bit simpler. It would look like this:
```elixir
def execute_command(commands, message) do
  # the rest of parse_command
  with :ok <- validate_arg_count(command, values),
          {:ok, values} <- parse_arg_values(command, values) do
    {:ok, apply(command.handler, values)}
  end
```
And `handle_message`'s body becomes only `execute_command(commands, message)`

This removes 2 privates functions by inlining them and makes the flow a bit easier to follow. What do you think?

test/mobius/core/command_test.exs
```diff
@@ -0,0 +1,81 @@
...
+        name: "hello",
+        args: [
+          foo: :integer,
+          bar: :string,
+          baz: :string
+        ],
+        handler: &command_handler/3
+      }
+
+      {:ok, command: command}
```
ZeLarpMaster: Can also do `[command: command]` btw

test/support/cog_stub.ex
```diff
@@ -0,0 +1,23 @@
...
+    send_to_test(message)
+  end
+
+  command "add", num1: :integer, num2: :integer do
+    send_to_test(num1 + num2)
+  end
+
+  defp send_to_test(message) do
+    :cog_test_process
+    |> Process.whereis()
```
ZeLarpMaster: `send` can receive the name of a locally registered process so you don't need to do `Process.whereis`

ZeLarpMaster (changes requested):
A few typos to be fixed, but overall looks very good! After those fixes it will be good to merge

lib/mobius/cog.ex
Blond11516: The Cog's `handle_info` is defined first from the macro expansion, so as long as the message matches the declaration, it handles it. I'll document this.

lib/mobius/core/command.ex
Blond11516: I agree with you, looking at the diff it does clean up the module nicely!

lib/mobius/cog.ex
Blond11516: I added some doc about this. Let me know what you think!

mix.lock
```diff
@@ -5,15 +5,20 @@
...
+  "makeup_elixir": {:hex, :makeup_elixir, "0.15.0", "98312c9f0d3730fde4049985a1105da5155bfe5c11e47bdc7406d88e01e4219b", [:mix], [{:makeup, "~> 1.0", [hex: :makeup, repo: "hexpm", optional: false]}, {:nimble_parsec, "~> 1.1", [hex: :nimble_parsec, repo: "hexpm", optional: false]}], "hexpm", "75ffa34ab1056b7e24844c90bfc62aaf6f3a37a15faa76b07bc5eba27e4a8b4a"},
   "metrics": {:hex, :metrics, "1.0.1", "25f094dea2cda98213cecc3aeff09e940299d950904393b2a29d191c346a8486", [:rebar3], [], "hexpm", "69b09adddc4f74a40716ae54d140f93beb0fb8978d8636eaded0c31b6f099f16"},
   "mime": {:hex, :mime, "1.4.0", "5066f14944b470286146047d2f73518cf5cca82f8e4815cf35d196b58cf07c47", [:mix], [], "hexpm", "75fa42c4228ea9a23f70f123c74ba7cece6a03b1fd474fe13f6a7a85c6ea4ff6"},
   "mimerl": {:hex, :mimerl, "1.2.0", "67e2d3f571088d5cfd3e550c383094b47159f3eee8ffa08e64106cdf5e981be3", [:rebar3], [], "hexpm", "f278585650aa581986264638ebf698f8bb19df297f66ad91b18910dfc6e19323"},
+  "nimble_parsec": {:hex, :nimble_parsec, "1.1.0", "3a6fca1550363552e54c216debb6a9e95bd8d32348938e13de5eda962c0d7f89", [:mix], [], "hexpm", "08eb32d66b706e913ff748f11694b17981c0b04a33ef470e33e11b3d3ac8f54b"},
```
ZeLarpMaster: I guess this sneaked in :P

ZeLarpMaster (dismissed):

ZeLarpMaster (dismissed):

ZeLarpMaster (approved):

Blond11516:
@all-contributors please add @Blond11516 for code, tests and documentation

PR #20, opened 2020-12-31
feat: Added connection ratelimiter service

ZeLarpMaster:
As [documented here](https://discord.com/developers/docs/topics/gateway#rate-limiting), Discord ratelimits the number of `Identify` it receives to a maximum of 1* per 5 seconds. To prevent network jitter from making this library accidentally break this limit, the library will instead wait until we receive a response from Discord before counting those 5 seconds. This makes the bot slower to identify on all shards if there is network jitter, but guarantees that we won't exceed the limit. This "wait for a response" is a design decision, not a requirement.

*The limit of *one* per 5 seconds is actually a concurrency limit specified in the [session start limit](https://discord.com/developers/docs/topics/gateway#session-start-limit-object) and can potentially be more than 1 though that's a maximum so a bot not using the maximum would still work properly, it would only identify very slowly if it has many shards.

TODO:
- [x] ~Support an arbitrary concurrency (currently only 1)~ Won't do now, this part isn't a top priority
- [x] Add a timeout to the ack
- [x] Use `Mobius.Services.ConnectionRatelimiter` where appropriate
- [x] Add tests for `Mobius.Services.ConnectionRatelimiter.Timed`
- [x] Add tests for the usages of `Mobius.Services.ConnectionRatelimiter`

### DIFF
--- a/lib/mobius/services/connection_ratelimiter/timed.ex
+++ b/lib/mobius/services/connection_ratelimiter/timed.ex
@@ -1,0 +1,89 @@
+defmodule Mobius.Services.ConnectionRatelimiter.Timed do
+  @moduledoc false
+
+  use GenServer
+
+  alias Mobius.Services.ConnectionRatelimiter
+
+  require Logger
+
+  @type state :: %{
+          queue: :queue.queue(pid),
+          current: pid | nil,
+          time_between_connections_ms: non_neg_integer,
+          timer_ref: reference | nil,
+          monitor_ref: reference | nil
+        }
+
+  # Server callbacks
+  @impl GenServer
+  @spec init(keyword) :: {:ok, state()}
+  def init(opts) do
+    state = %{
+      queue: :queue.new(),
+      current: nil,
+      time_between_connections_ms: Keyword.fetch!(opts, :time_between_connections_ms),
+      timer_ref: nil,
+      monitor_ref: nil
+    }
+
+    {:ok, state}
+  end
+
+  @impl GenServer
+  def handle_call({:connect, pid}, _from, %{current: nil} = state) do
+    ref = unblock_process(pid)
+    {:reply, :ok, %{state | current: pid, monitor_ref: ref}}
+  end
+
+  def handle_call({:connect, pid}, _from, state) do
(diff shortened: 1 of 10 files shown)


Blond11516 (commented):
Looking good so far

lib/mobius/services/connection_ratelimiter/timed.ex
```diff
@@ -0,0 +1,117 @@
...
+    end
+  end
+
+  @impl GenServer
+  def handle_info(:ack_timeout, state) do
+    GenServer.cast(__MODULE__, {:connect_ack, state.current})
+    {:noreply, state}
+  end
+
+  def handle_info(:unblock_next, state) do
```
Blond11516, line 92:
```suggestion
  @impl GenServer
  def handle_info(:unblock_next, state) do
```

test/mobius/services/connection_ratelimiter/timed_test.exs
```diff
@@ -0,0 +1,64 @@
...
+      assert_received :callback_called
+      Timed.ack_connected()
+      Timed.wait_until_can_connect(make_callback())
+
+      assert_callback_called(@connection_delay)
+    end
+
+    test "executes callbacks a delay after the previous ack timed out" do
+      Timed.wait_until_can_connect(make_callback())
+      assert_received :callback_called
```
Blond11516: Why not `assert_callback_called(0)`?

Blond11516 (dismissed):

lib/mobius/services/connection_ratelimiter/timed.ex
ZeLarpMaster: The `@impl` applies to `handle_info/2` which is already defined one function above (for `:ack_timeout`) so adding it here again wouldn't do anything and would probably prompt a warning from the linter or the compiler

test/mobius/services/connection_ratelimiter/timed_test.exs
ZeLarpMaster: Oh I just missed those while refactoring, I'll change that

Blond11516 (approved):

PR #25, opened 2021-01-10
feat: Support intents

ZeLarpMaster:
Documentation on intents: https://discord.com/developers/docs/topics/gateway#gateway-intents

TODO:
- [x] Warn about privileged intents if socket closes with 4014 (with test)
- [x] Make the identify use the intent (with tests)
- [x] Store the intents somewhere accessible

### DIFF
--- a/lib/mobius/core/bitflags.ex
+++ b/lib/mobius/core/bitflags.ex
@@ -33,14 +33,7 @@ defmodule Mobius.Core.Bitflags do
       0b1010
   """
   @spec parse_bitflags(integer, [arg | nil], MapSet.t(arg)) :: MapSet.t(arg) when arg: var
-  def parse_bitflags(num, flags, out \\ MapSet.new())
-  def parse_bitflags(_num, [], out), do: out
-  def parse_bitflags(0, _flags, out), do: out
-
-  def parse_bitflags(num, [flag | flags], out) when is_integer(num) do
-    out = if (num &&& 1) == 1 and flag != nil, do: MapSet.put(out, flag), else: out
-    parse_bitflags(num >>> 1, flags, out)
-  end
+  def parse_bitflags(num, flags), do: parse_bitflags(num, flags, MapSet.new())
 
   @doc """
   Converts a list of flags into an integer of bitflags
@@ -76,5 +69,19 @@ defmodule Mobius.Core.Bitflags do
     |> Stream.filter(fn {flag, _index} -> flag in input end)
     |> Stream.map(fn {_flag, index} -> index end)
     |> Enum.reduce(0, &Bitwise.bor/2)
+  end
+
+  defp parse_bitflags(_num, [], out), do: out
+  defp parse_bitflags(0, _flags, out), do: out
+
+  defp parse_bitflags(num, [flag | flags], out) when is_integer(num) do
+    out =
+      if (num &&& 1) == 1 and flag != nil do
+        MapSet.put(out, flag)
+      else
+        out
+      end
+
+    parse_bitflags(num >>> 1, flags, out)
   end
 end
--- a/lib/mobius/core/bitflags.ex
+++ b/lib/mobius/core/bitflags.ex
@@ -38,7 +38,7 @@ defmodule Mobius.Core.Bitflags do
   @doc """
   Converts a list of flags into an integer of bitflags
 
-  Foreach flag, if it can be found in the input, its associated bit will be 1 in the result.
+  For each flag, if it can be found in the input, its associated bit will be 1 in the result.
   Starting from the least significant bit.
 
   This function is the inverse of `parse_bitflags/2`
--- a/lib/mobius/core/bitflags.ex
+++ b/lib/mobius/core/bitflags.ex
@@ -1,0 +1,80 @@
+defmodule Mobius.Core.Bitflags do
+  @moduledoc false
+
+  import Bitwise
+
+  @doc """
+  Converts bitflags into a MapSet of flag values
+
+  Values from the list are included starting from the least significant bit.
+  Parsing is stopped and the current result is returned when the list of things is exhausted.
+
+  This function is the inverse of `create_bitflags/2`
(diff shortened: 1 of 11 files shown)


lib/mobius/core/bitflags.ex
```diff
@@ -0,0 +1,80 @@
...
+
+      iex> parse_bitflags(0b11111111, [])
+      #MapSet<[]>
+
+      iex> flags = [:a, :b, :c, :d]
+      iex> parse_bitflags(0b1010, flags) |> create_bitflags(flags)
+      0b1010
+  """
+  @spec parse_bitflags(integer, [arg | nil], MapSet.t(arg)) :: MapSet.t(arg) when arg: var
+  def parse_bitflags(num, flags, out \\ MapSet.new())
```
Blond11516: When writing recursive functions like this I usually prefer to have a public function without an accumulator call a private implementation with an accumulator. This makes the intent of the function clearer imo.

Of course, disregard this if you _do_ want users to be able to pass a non-empty initial accumulator.

lib/mobius/core/bitflags.ex
```diff
@@ -0,0 +1,80 @@
...
+  def parse_bitflags(_num, [], out), do: out
+  def parse_bitflags(0, _flags, out), do: out
+
+  def parse_bitflags(num, [flag | flags], out) when is_integer(num) do
+    out = if (num &&& 1) == 1 and flag != nil, do: MapSet.put(out, flag), else: out
```
Blond11516: I usually prefer multiline if statements, especially when they have an `else` block. I feel like this line is a little dense.

lib/mobius/core/bitflags.ex
```diff
@@ -0,0 +1,80 @@
...
+    parse_bitflags(num >>> 1, flags, out)
+  end
+
+  @doc """
+  Converts a list of flags into an integer of bitflags
+
+  Foreach flag, if it can be found in the input, its associated bit will be 1 in the result.
```
Blond11516:
```suggestion
  For each flag, if it can be found in the input, its associated bit will be 1 in the result.
```

lib/mobius/core/intents.ex
```diff
@@ -0,0 +1,119 @@
...
+
+  for {event_name, required_intents} <- @event_intents do
+    if required_intents == [] do
+      def has_intent_for_event?(unquote(event_name), _intents), do: true
+    else
+      def has_intent_for_event?(unquote(event_name), intents) do
+        Enum.any?(unquote(required_intents), &(&1 in intents))
+      end
+    end
+  end
```
Blond11516: Is this really necessary? Couldn't we do something similar at runtime?

lib/mobius/core/intents.ex
```diff
@@ -0,0 +1,119 @@
...
+    message_reaction_remove: [:guild_message_reactions, :direct_message_reactions],
+    message_reaction_remove_all: [:guild_message_reactions, :direct_message_reactions],
+    message_reaction_remove_emoji: [:guild_message_reactions, :direct_message_reactions],
+    presence_update: [:guild_presences],
+    typing_start: [:guild_message_typing, :direct_message_typing],
+    user_update: [],
+    voice_state_update: [:guild_voice_states],
+    voice_server_update: [],
+    webhooks_update: [:guild_webhooks]
+  ]
```
Blond11516: Why not define all of these as maps/map sets? All usages either do that or iterate over them.

lib/mobius/core/intents.ex
```diff
@@ -0,0 +1,119 @@
...
+  """
+  @spec events_for_intents(t()) :: list(atom)
+  def events_for_intents(intents) do
+    @event_intents
+    |> Enum.map(fn {name, _} -> name end)
+    |> Enum.filter(&has_intent_for_event?(&1, intents))
+  end
+
+  @doc "Returns true if the event can be received with the given intents. Returns false otherwise"
+  @spec has_intent_for_event?(atom, list(atom)) :: boolean
```
Blond11516: Why not take a `MapSet` here? All other functions do.

Also, without having looked at usages, moving the intents to the first parameter should make this more pipe-friendly.

lib/mobius/core/intents.ex
Blond11516: Also, per my previous comment about the attributes, I think this could be done trivially and very efficiently at runtime if `@event_events` was converted to a `Map`.

```elixir
Map.get!(event_name)
|> Enum.any?(&(&1 in intents))
```

Blond11516 (dismissed):
Some nice modules in there! A few comments for cleaning up `BitFlags` and `Intents`, but the overall flow is pretty good!

lib/mobius/core/intents.ex
ZeLarpMaster: `@intents` needs to be ordered for the bitflags (it determines which bit is for which intent) which is why that one is a list.

Also they're all lists because for some reason sets at compile time cause this error:
image

`@event_intents` could probably be a map though, I'll try that out

lib/mobius/core/intents.ex
ZeLarpMaster: I think the type was forgotten after refactoring things.

I honestly don't think either ordering of parameters would be helpful in a pipeline since it returns a boolean. It's more likely gonna be used like `if has_intent_for_event?(event, intents) do` than some sort of pipeline

lib/mobius/core/intents.ex
Blond11516: Oh right, I guess the comment saying the ordering is important didn't make it through to my brain :sweat_smile:

lib/mobius/core/intents.ex
Blond11516: Yeah I agree it probably doesn't matter too much in this case, it's just a general rule of thumb I try to follow :stuck_out_tongue:

Blond11516 (approved):

PR #26, opened 2021-01-10
feat: Handling token revokes

ZeLarpMaster:
4003 is sent immediately when the token is revoked, so forcing a "no-reconnect" policy on it means we stop receiving events from Discord which should stop most interactions happening. There's still a possibility of a background task doing http requests or a delayed interaction happening, but it should be fine for most cases. This solution is only temporary anyway. Once the token is moved elsewhere we'll be able to implement a more proper solution.

Shutting down BEAM was an option until I thought about the impact it would have on tests. It makes it complicated to test and this is a temporary solution anyway so it wasn't worth the effort.

I also changed a bit of the retry logic on http requests so requests whose result won't change after a retry aren't retried.


Blond11516 (approved):
Very simple, I love it! :wink:

PR #37, opened 2021-01-16
feat: Adding Discord data models and parsers

### LINKED ISSUE #28: Parsing Discord models into structs
This issue is for tracking which models have been implemented. Refer to the documentation for information on what they contain: https://discord.com/developers/docs/intro (you may need to look around to find a specific model, but they were all in there at the moment this was written; feel free to ask if you can't find something)

- [x] Channel
- [x] Message
- [x] Permissions
- [x] User
- [x] Snowflake
- [x] ISO8601 timestamp
- [x] Member
- [x] Role
- [x] ChannelMention
- [x] Attachment
- [x] Embed
- [x] Reaction
- [x] MessageActivity
- [x] MessageApplication
- [x] MessageReference
- [x] Sticker
- [x] Emoji
- [x] Guild
- [x] VoiceState
- [x] Presence
- [x] GuildPreview
- [x] Ban
- [x] GuildWidget
- [x] Integration
- [x] Invite
- [x] InviteMetadata
- [x] GuildTemplate
- [x] VoiceRegion
- [x] Webhook
- [x] Team
- [x] TeamMember
- [x] Application
- [x] GatewayBot
- [x] SessionStartLimit

ZeLarpMaster:
Implements a bunch of models listed in #28

There's a total of 34 models to implement with increasingly more complexity in them. It is therefore important to discuss and review the API of models and their implementation to seek the best approach possible before implementing all the other models.

The general idea behind the current architecture is attempting to avoid non-breaking changes on Discord's side to affect users in a negative way. So new fields should simply be ignored. Everything else should probably (up to discussion) fail fast with clear messages to easily detect mistakes or breaking changes.

All fields are currently considered as optional and will use the value `nil` when it isn't there. This is honestly just laziness and to avoid making the implementation too complex.

### DIFF
--- a/lib/mobius/models/gateway_bot.ex
+++ b/lib/mobius/models/gateway_bot.ex
@@ -1,0 +1,42 @@
+defmodule Mobius.Models.GatewayBot do
+  @moduledoc """
+  Struct for the response of a GET /gateway/bot request
+
+  Related documentation: https://discord.com/developers/docs/topics/gateway#get-gateway-bot
+  """
+
+  import Mobius.Models.Utils
+
+  alias Mobius.Models.SessionLimit
+
+  defstruct [:url, :shards, :session_limit]
+
+  @type t :: %__MODULE__{
+          url: String.t(),
+          shards: non_neg_integer(),
+          session_limit: SessionLimit.t()
+        }
+
+  @doc """
+  Parses the given term into a `t:t()` if possible; returns nil otherwise
+
+  ## Examples
+
+      iex> alias Mobius.Models.{GatewayBot, SessionLimit}
+      iex> parse("not a map")
+      nil
+      iex> parse(%{})
+      %GatewayBot{}
+      iex> parse(%{"url" => "wss://something", "shards" => 1, "session_limit" => %{}})
+      %GatewayBot{url: "wss://something", shards: 1, session_limit: %SessionLimit{}}
+  """
+  @spec parse(any) :: t() | nil
+  def parse(map) when is_map(map) do
+    %__MODULE__{}
+    |> add_field(map, "url", :url)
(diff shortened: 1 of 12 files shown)


lib/mobius/models/gateway_bot.ex
```diff
@@ -0,0 +1,42 @@
...
+      %GatewayBot{}
+      iex> parse(%{"url" => "wss://something", "shards" => 1, "session_start_limit" => %{}})
+      %GatewayBot{url: "wss://something", shards: 1, session_limit: %SessionLimit{}}
+  """
+  @spec parse(any) :: t() | nil
+  def parse(map) when is_map(map) do
+    %__MODULE__{}
+    |> add_field(map, "url", :url)
+    |> add_field(map, "shards", :shards)
+    |> add_field(map, "session_start_limit", :session_limit, &SessionLimit.parse/1)
```
Blond11516: Is it really worth it to change the name of the field? I feel like it might create some unnecessary friction between Mobius and the Discord docs.

If not, I'd suggest changing `add_field` to remove either the 3rd or 4th argument and convert accordingly as necessary.

Blond11516 (dismissed):
Really liking this! I think this gives a good mix of flexibility, robustness and future-proofing. :+1:

lib/mobius/models/gateway_bot.ex
ZeLarpMaster: You're right, it might add some friction. I'll change it and make add_field automatically convert from atom to string instead.

It was mostly because I hate when mix format splits one-liners into multiple lines just because they're too long, but by making add_field convert instead, it will be shorter and won't happen anyway. So yeah, very good idea! I love it

Blond11516 (approved):

PR #38, opened 2021-01-17
feat(models): Application and its dependencies

ZeLarpMaster:
Same thing as #37 but for Application which can be found here: https://discord.com/developers/docs/topics/oauth2#application-object

TODO:
- [x] Tests for `Mobius.Models.Team`
- [x] Tests for `Mobius.Models.TeamMember`
- [x] Tests for `Mobius.Models.Application`
- [x] Tests for `Mobius.Models.User`
- [x] Usage


lib/mobius/core/bitflags.ex
```diff
@@ -16,6 +17,9 @@ defmodule Mobius.Core.Bitflags do
       iex> parse_bitflags(0b01, [:a, :b])
       #MapSet<[:a]>
 
+      iex> parse_bitflags(0b101, [:a, nil, :b])
+      #MapSet<[:a, :b]>
```
Blond11516: Wouldn't this test still be true with any value other than `nil`?

lib/mobius/core/bitflags.ex
```diff
@@ -48,6 +53,9 @@ defmodule Mobius.Core.Bitflags do
       iex> create_bitflags(MapSet.new([:a, :b]), [:a, :b])
       0b11
 
+      iex> create_bitflags(MapSet.new([:a, :b]), [nil, :a, :b])
+      0b110
```
Blond11516, line 51: Same comment as above.

lib/mobius/models/snowflake.ex
```diff
@@ -0,0 +1,73 @@
...
+
+  import Bitwise
+
+  alias Mobius.Models.Utils
+
+  @type t :: pos_integer()
+
+  @discord_epoch 1_420_070_400_000
+
+  @doc "Calls `Mobius.Models.Utils.parse_integer/1`"
```
Blond11516: Ok but why would I want to do this? This doc isn't really helpful...

lib/mobius/models/user.ex
```diff
@@ -0,0 +1,118 @@
...
+          | :partnered_server_owner
+          | :hypesquad_events
+          | :bug_hunter_level_1
+          | :house_bravery
+          | :house_brilliance
+          | :house_balance
+          | :early_supporter
+          | :team_user
+          | :system
+          | :bug_hunter_level_1
```
Blond11516:
```suggestion
          | :bug_hunter_level_2
```

lib/mobius/models/utils.ex
```diff
@@ -19,4 +19,19 @@ defmodule Mobius.Models.Utils do
 
     struct!(struct, [{struct_key, value}])
   end
+
+  @spec parse_list(any, (any -> output)) :: [output] | nil when output: var
```
Blond11516, line 23:
```suggestion
  @spec parse_list(input, (Enum.t(input) -> output)) :: [output] | nil when output: var, input: var
```

Not quite sure this is right, but I'm pretty sure we could a constraint for the input.

test/support/test_utils.ex
```diff
@@ -36,4 +36,19 @@ defmodule Mobius.TestUtils do
...
+  Asserts that a struct's field contains the expected value and returns the struct
+
+  ## Examples
+
+      iex> my_struct = %MyStruct{field: "value"}
+      iex> check_field(my_struct, :field, "value") == my_struct
+      true
+  """
+  @spec check_field(arg, atom, any) :: arg when arg: struct
+  def check_field(struct, field, expected_value) do
```
Blond11516: nit: I feel like `assert_field` would be more accurate. I think most people would assume that it does assert, but I'd rather be clear about it.

test/support/generators.ex
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    merge_opts(defaults, opts)
+  end
+
+  defp merge_opts(defaults, opts) do
+    opts_map =
+      opts
+      |> Enum.map(fn {key, value} -> {Atom.to_string(key), value} end)
+      |> Map.new()
```
Blond11516:
```suggestion
    opts_map = Map.new(opts, fn {key, value} -> {Atom.to_string(key), value} end)
```

Blond11516 (commented):
Gotta admit, I didn't check the fields against the Discord docs, but the code itself mostly looks good :+1:

lib/mobius/core/bitflags.ex
ZeLarpMaster: It's true for any integer where the 3 least significant bits are `0b101` or `0b111`

Unless you mean the list of flags?

lib/mobius/models/snowflake.ex
ZeLarpMaster: Fair point

lib/mobius/models/utils.ex
ZeLarpMaster: We could if the function raised an error when the input isn't a list, but this function is perfectly valid for non-lists, it just returns nil

Also the input of the function isn't `Enum.t`, it's the individual elements of the list. The success typing of this function is actually `@spec parse_list([input], (input -> output)) :: [output] when input: var, output: var` which is what I had there until I added the nil return

test/support/test_utils.ex
ZeLarpMaster: Good point, it felt weird to me to call it assert though since asserts don't return anything and this function does, but you're right that it might be clearer

lib/mobius/core/bitflags.ex
Blond11516: `0b111` was what I was going for. The current test doesn't showcase the effect of `nil` because the corresponding bit is `0`, so it doesn't matter if the 2nd value is `nil`, it's not gonna be included anyway.

lib/mobius/models/utils.ex
Blond11516: You're right, forgot about the nil return.

Blond11516 (approved):

lib/mobius/core/bitflags.ex
ZeLarpMaster: Ah I see, I changed the test to make it 0b111 instead 👍

lib/mobius/core/bitflags.ex
ZeLarpMaster: Well create_bitflags has no way of knowing whether to put a 0 or a 1 for `nil` values so it puts a 0. You're right that it doesn't test it, but it's impossible to test it. This test simply shows how to use it.

PR #45, opened 2021-01-23
feat(models): Added Permissions

ZeLarpMaster:
Same thing as #37 but for Permissions which can be found here: https://discord.com/developers/docs/topics/permissions

**This PR depends on #43**


lib/mobius/models/permissions.ex
```diff
@@ -0,0 +1,96 @@
...
+          | :use_vad
+          | :change_nickname
+          | :manage_nicknames
+          | :manage_roles
+          | :manage_webhooks
+          | :manage_emojis
+
+  @type t :: MapSet.t(permission())
+
+  @doc "Returns all permissions possible"
```
Blond11516:
```suggestion
  @doc "Returns all existing permissions"
```

Blond11516 (dismissed):

Blond11516 (approved):

PR #51, opened 2021-01-29
feat: Preventing requests to be made with invalid tokens

### LINKED ISSUE #27: Stop http requests when the token is invalid
As explained in #26, token revokes currently shutdown the shard, but don't prevent http requests from being made. This causes a problem where requests might be made even after the last received event (for example if there was a delayed event or a background task).

Ideally, as soon as we know the token is invalid, we should prevent any further http requests.

Related documentation: 401 at https://discord.com/developers/docs/topics/rate-limits#invalid-request-limit

ZeLarpMaster:
Uses `:fuse` to prevent requests as soon as the token is known to be invalid

This approach currently requires a single 401 to be received before stopping further requests. I'm wondering whether it would be worth pursuing ways to not require that 401. Specifically, when the token is revoked, the shard is shutdown instantly with a specific socket code which could potentially be used to melt the fuse and prevent requests.

Closes #27


Blond11516 (approved):
Not too familiar with fuse, but nothing jumps out to me in there.

PR #52, opened 2021-01-30
feat: Added a random sleep after failing to resume

### LINKED ISSUE #36: feat: Sleep randomly after a failed resume
The [documentation on resuming](https://discord.com/developers/docs/topics/gateway#resuming) explains:
> It's also possible that your client cannot reconnect in time to resume, in which case the client will receive a Opcode 9 Invalid Session and is expected to __wait a random amount of time—between 1 and 5 seconds—then send a fresh Opcode 2 Identify__

This part is mentioned in the 2 TODOs left in shard.ex. Note that those comments are out of date and a proper solution may not involve either of those lines.

ZeLarpMaster:
I had no idea how to test this so I didn't, if you have any suggestions I'm open to it

Closes #36

### DIFF
--- a/config/config_test.exs
+++ b/config/config_test.exs
@@ -5,6 +5,7 @@ config :ex_unit,
   refute_receive_timeout: 50
 
 config :mobius,
+  resuming_sleep_time_ms: 50,
   ratelimiter_impl: Mobius.Stubs.CommandsRatelimiter,
   connection_ratelimiter_impl: Mobius.Stubs.ConnectionRatelimiter,
   socket_impl: Mobius.Stubs.Socket,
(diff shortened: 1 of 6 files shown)


Blond11516 (commented):
Code looks good, but I'd really like to see some tests. The test would have to look something like the following:
1. Mock a failed connection
2. Make the bot send a message to detect the failed connection
3. Mock an invalid session
4. Wait `resuming_sleep_time_ms` minus some security buffer
4. Make sure we haven't received identify
5. Wait for the duration of the security buffer
6. Make sure we have received identify

No idea how that would translate to code, or if it's event possible to do all this with our current setup...

ZeLarpMaster:
That's a very good idea, I didn't think about the identify which is supposed to arrive right after the 2nd attempt to reconnect. I'll try to make it work :+1:

Blond11516 (approved):

PR #53, opened 2021-02-02
feat(models): Added Embed

ZeLarpMaster:
Same thing as #37 but for Embed

Note: This also includes the renaming of `Mobius.Models.DateTime` --> `Mobius.Models.Timestamp` because it conflicted with `Elixir.DateTime` which made it annoying to use.

### DIFF
--- a/lib/mobius/models/embed.ex
+++ b/lib/mobius/models/embed.ex
@@ -1,0 +1,82 @@
+defmodule Mobius.Models.Embed do
+  @moduledoc """
+  Struct for Discord's Embed
+
+  Related documentation:
+  https://discord.com/developers/docs/resources/channel#embed-object
+  """
+
+  import Mobius.Models.Utils
+
+  alias Mobius.Models.Timestamp
+
+  defstruct [
+    :title,
+    :type,
+    :description,
+    :url,
+    :timestamp,
+    :color,
+    :footer,
+    :image,
+    :thumbnail,
+    :video,
+    :provider,
+    :author,
+    :fields
+  ]
+
+  @type type ::
+          :image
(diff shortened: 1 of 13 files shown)


lib/mobius/models/embed.ex
```diff
@@ -0,0 +1,82 @@
...
+    :author,
+    :fields
+  ]
+
+  @type type ::
+          :image
+          | :video
+          | :gifv
+          | :article
+          | :link
```
Blond11516: No `:rich`?

lib/mobius/models/embed.ex
```diff
@@ -0,0 +1,82 @@
...
+
+  def parse(_), do: nil
+
+  defp parse_type("rich"), do: :rich
+  defp parse_type("image"), do: :image
+  defp parse_type("video"), do: :video
+  defp parse_type("gifv"), do: :gifv
+  defp parse_type("article"), do: :article
+  defp parse_type("link"), do: :link
+  defp parse_type(_), do: nil
```
Blond11516, line 80: If `String.to_existing_atom/1` returned some kind of error I would've suggested to use it but `rescue` would be kinda ugly :/

lib/mobius/models/embed.ex
ZeLarpMaster: Derp

lib/mobius/models/embed.ex
ZeLarpMaster: Yeah, I prefer this, it's not that bothersome and makes it very explicit

Blond11516 (approved):

PR #55, opened 2021-02-03
feat: Added a POC cache system

ZeLarpMaster:
The current idea is to use Cachex as the provider and to cache everything we see coming through. I doubt this is the way to go about it though because, for example, a partial user (only id, username, discriminator, and avatar) would replace a full user (13 fields) if it comes after. Therefore some kind of merging system should be put in place. I'm not sure whether we want to cache only the events which give the most information or cache everything with a merge.

Caching everything with merges would be more accurate since we'd constantly update the information we have in cache, but it also involves a bunch of work. Whereas when caching only on events which give the most information, we do less work for less accurate information.


Blond11516 (commented):
This looks acceptable as a first draft, but I have a few issues with the current implementation:
1. I feel like the way this is going, caching is going to be very painful to maintain. Surely there's a better way than manually picking and caching data in each event?
2. I'm not a very big fan of building a caching system that doesn't guarantee accuracy of the data. I'm sure this is fine for a number of use cases, but for a general purpose framework I think it could be a big deal-breaker for some people.
3. Do we really need this right now? I'm not aware of how Discord bots usually perform, but I don't necessarily feel like a caching system is MVP material, especially given the complexity of building such a system properly.

ZeLarpMaster:
Before anything, you're raising very good points and made me rethink a lot about the direction of the caches.

[Caching is expected by Discord](https://discord.com/developers/docs/topics/gateway#tracking-state) and from experience the caches are used by all sorts of cogs. I know we can't guarantee accuracy of the data, but, even if our caching system was perfect, [Discord events aren't guaranteed either](https://discord.com/developers/docs/reference#consistency) so we the cache wouldn't be accurate anyway. The goal of the cache is only to avoid doing unnecessary requests.

About your first point, I agree this is kind of dumb. Reading back in the docs, I see that Discord doesn't expect us to grab user data from literally every single place it could come from, so making those changes I think only two user cache usages will be left. At the end of the day we kind of need to handpick everywhere relevant to cache since there's no generic way to cache "anything". Discord doesn't tell us "hey, add this to your cache" or "hey update this in your cache", they only tell us "hey this message happened" and we have to manually cache the message and the information we receive with it. This also gives me a pretty good idea of a direction for the caches: following the Discord events' naming scheme and caching all of it. If you look at the event names, they're pretty much all named "{Model} {CUD action}" with little variation. I think this should map pretty easily to caches and their usage. Hopefully this will look better for maintenance too.

About the third point, I do think it's a valuable thing to have on release. All other Discord libs have some sort of cache and I know for a fact some of them will be useful for the builtin cogs. One example is a userinfo command, it lists out the roles a user has, the date they joined the server, and things like that. But you've made me realize all the caches I planned to do are actually available on API calls. This makes me wonder whether a caching solution around the API would be better. Maybe we could use Cachex to prevent calls to the API, but I feel like this could force users to use outdated data 🤔 Food for thought I guess

Blond11516:
Reading the "Tracking State" section of the docs (your first link) I get the understanding that caching should ideally be custom tailored to each bot/cog's needs? Is this something that other frameworks/libraries usually do?

The point of sticking to event names is great. I agree this should make managing the caching system much simpler!

The idea of caching API calls is also interesting, but I agree that data might get outdated very quickly since discord is very "push-based" in many cases.

lib/mobius/services/model_cache.ex
```diff
@@ -43,46 +43,56 @@ defmodule Mobius.Services.ModelCache do
...
-    users = Enum.map(users || [], fn user -> {user["id"], user} end)
-    Cachex.put_many(__MODULE__.User, users)
+  defp cache_member(member) do
+    user = member["user"]
+    cache_user(user)
+    Cachex.put(__MODULE__.Member, {member["guild_id"], user["id"]}, member)
   end
 
-  defp members_to_users(members), do: Enum.map(members || [], fn member -> member["user"] end)
+  defp uncache_member(%{"guild_id" => guild_id, "user" => %{"id" => id}}) do
```
Blond11516: How about `invalidate_member`?

ZeLarpMaster: Sounds much better indeed

ZeLarpMaster:
I haven't seen any bot do custom caching, but I have only looked at 4 implementations out of dozens. But one thing which we can do, is if cogs declare the [intents](https://discord.com/developers/docs/topics/gateway#gateway-intents) they have, we can make the bot only use the necessary intents which subsequently means we only receive events for those intents. This means we receive less information to cache overall which makes it so we have a kinda custom tailored cache. All the caches would still be there, but they'd be empty all the time. Later on we could add some checks for intents so users know they don't have the intents for a specific cache.

lib/mobius/actions/user.ex
```diff
@@ -0,0 +1,26 @@
...
+  [Discord documentation](https://discord.com/developers/docs/reference#consistency),
+  events may never be sent to a client (this library).
+  This means the cache may not contain things it should contain
+  or may be out of date if an update event isn't sent (or received).
+
+  If being up to date is very important to you,
+  you may request up to date information with `get_user/1`
+  """
+  @spec get_cached_user(Snowflake.t()) :: User.t() | nil
+  def get_cached_user(id), do: ModelCache.get(id, ModelCache.User)
```
Blond11516, line 25: Do we really need a way to explicitly query the cache without automatically falling back to a network request? From you experience are there really that many cases where we'd want to get some data only if it's cached?

lib/mobius/services/model_cache.ex
```diff
@@ -0,0 +1,111 @@
...
+    value
+  end
+
+  @spec list(cache()) :: [any]
+  def list(cache) do
+    {:ok, stream} = Cachex.stream(cache, Cachex.Query.create(true, :value))
+    Enum.to_list(stream)
+  end
+
+  @spec cache_event(Event.name(), any) :: any
```
Blond11516, line 66: This might be a good case for a little function generation macro goodness. I'm thinking looping over a list of tuples like `{event_name, event_cache_handler}` to create the corresponding function heads (with a static catch-all head for non-handled events).

Higher chance of error, but much less boilerplate to maintain.

Blond11516 (commented):
Looks good as a first cache implementation. Only a few small comments, and missing tests :)

lib/mobius/actions/user.ex
ZeLarpMaster: I prefer giving users the option rather than forcing them on something. If someone has to get information on a batch of users at once, it would very very inefficient to do a call for each one of them. Or maybe a raid happens and you wanna know when they all joined so you need to access all of those Members. I think there's many instances where having the option of only getting the cache is good.

Also, getting a user with a request is gonna be `get_user(id)` (or fetch?) and `get_cached_user/1` returns a nil if not found, so someone who wants to fallback to a network call only needs to do `get_cached_user(id) || get_user(id)` which could eventually be in a function like `get_or_fetch_user(id)` if necessary.

lib/mobius/services/model_cache.ex
ZeLarpMaster: Would it really do much though? There isn't much boilerplate outside the definition of a function and the name of the event and all of these functions are gonna be different anyway so it would end up looking very similar in terms of code unless I'm misunderstanding your suggestion 🤔

Blond11516: Thinking about it again, you're right that a macro probably wouldn't help much. The main thing I'd like to improve is the visibility of the functions that _are_ implemented (as in, don't return `nil`). There are a few with single line bodies that are kind of getting lost in a sea of `, do: nil`.

One way to fix this would simply be to reorganize the function heads to group the implemented functions together, but I'm not sure why we even need all those heads to explicitly return `nil`. Couldn't we just have a catch-all head like `cache_event(_, _), do: nil`?

If you really think there is value in matching the exact list of event names that could also be done in a function guard. In fact, we already have `Mobius.Core.Event.is_event_name?/1` which does exactly that.

ZeLarpMaster: The only reason the whole list is there right now is because I wanted an easy way to know which events were available without needing to change files every 2 minutes. I can for sure remove all the unused ones for now though.

Blond11516 (dismissed):

lib/mobius/services/model_cache.ex
```diff
@@ -0,0 +1,108 @@
...
+  def get(key, cache) do
+    {:ok, value} = Cachex.get(cache, key)
+    value
+  end
+
+  @doc "Lists all the values in a given cache"
+  @spec list(cache()) :: [any]
+  def list(cache) do
+    {:ok, stream} = Cachex.stream(cache, Cachex.Query.create(true, :value))
+    Enum.to_list(stream)
```
ZeLarpMaster: Thoughts on this being a list instead of the stream? I don't remember why I did that and it feels a bit dumb now that I think about it 🤔

Blond11516: Lists are more "conventional I guess"? Both are probably fine considering that it's likely going to be consumed by `Enum` anyway...

lib/mobius/services/model_cache.ex
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  @doc "Clears all caches. Only for usage in tests."
```
Blond11516: What if I really want to clear my cache for some weird reason? If it's really only for tests, I don't really like the idea of this being in here.

Blond11516 (dismissed):

lib/mobius/services/model_cache.ex
ZeLarpMaster: Fair enough. I mostly want to avoid someone using it unknowingly, but I guess it already does that by being a service-only function (aka a function which isn't also in a `Mobius.Actions` module). I really don't think this is something which should ever be made available to users though since it would break a lot of the bot's capacity to prevent errors. Maybe someone could really want to clear the whole cache, but they could just restart the bot to refresh it instead.

Maybe I was just lazy and didn't want to restart the whole app between every cache test, but I do think this has value for tests.

tl;dr: I'd remove `Only for usage in tests.`, but keep it service-only. If you disagree with that I guess we can discuss this further

Blond11516: Your point of restarting the bot instead of clearing the cache is very fair!

Just changing the doc is fine by me; I don't really care about the function itself, but seeing that kind of comments just rubs me the wrong way :stuck_out_tongue:

test/mobius/services/model_cache_test.exs
```diff
@@ -102,7 +102,7 @@ defmodule Mobius.Services.ModelCacheTest do
     end
 
     test "returns an empty list if nothing cached" do
-      assert [] == ModelCache.list(ModelCache.User)
+      assert [] == Enum.to_list(ModelCache.list(ModelCache.User))
```
Blond11516:
```suggestion
      assert Enum.empty(ModelCache.list(ModelCache.User))
```

Blond11516 (dismissed):

Blond11516 (approved):

PR #60, opened 2021-02-21
feat(models): Added VoiceState

ZeLarpMaster:
Same thing as #37 but for VoiceState and Presence (which includes Activity)

Presence (and Activity) included some pretty niche fields which would be very annoying to parse into individual structs so I've made the choice of keeping those as maps with strings as keys. If you disagree with this decision, do tell me.

### DIFF
--- a/lib/mobius/models/activity.ex
+++ b/lib/mobius/models/activity.ex
@@ -1,0 +1,91 @@
+defmodule Mobius.Models.Activity do
+  @moduledoc """
+  Struct for Discord's Activity
+
+  Related documentation:
+  https://discord.com/developers/docs/topics/gateway#activity-object
+  """
+
+  import Mobius.Models.Utils
+
+  alias Mobius.Models.Snowflake
+
+  defstruct [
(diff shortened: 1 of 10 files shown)


Blond11516 (commented):
I think there would be _some_ value in defining structs for all fields, if only for consistency's sake, but it can definitely wait for later.

Another option would be to define they keys those maps expect in the typespecs instead of using generic maps.

ZeLarpMaster:
> Another option would be to define they keys those maps expect in the typespecs instead of using generic maps.

I thought about that and tried it, but typespecs really don't like strings as keys in maps. The only way it can work is something like `%{required(String.t()) => type_of_field1() | type_of_field2() | type_of_field3()}` which is absolutely not helpful. It would work if the keys were atoms, but that kinda goes against the point of having a map instead of a struct

Blond11516 (approved):

Blond11516:
Got it. It kinda sucks but I'm not too surprised it doesn't work...

Issue #29, opened 2021-01-11
Implement all relevant HTTP endpoints


ZeLarpMaster:
Refer to [this spreadsheet](https://docs.google.com/spreadsheets/d/19-tAK4h8jS68vvama2HYTUFsrE1RZSSq3WnE1lnccHs/edit) for the status of each individual http endpoint (the background color of a row is green for done, red for don't do, white for available). If you implement one, please mention a maintainer in your PR so they can update the spreadsheet once your PR is merged.

These implementations should be strictly about making the HTTP call and transforming data for it. Any checks for permissions or limits should be in the `Mobius.Actions` related function and not in `Mobius.Rest`.

PR #62, opened 2021-02-27
feat(actions): Implementing send message

ZeLarpMaster:
Adds the `send_message` action. A future PR will add `send_file`

It also includes a little bit of refactoring (moving tests in status_test.exs into the describe; removing `create_token` from fixtures) and fixing tests (self_refill_test.exs somehow broke on me even though there's a one in 16^8 chance that the same bucket is used; I don't know whether that's really what happened, but 🤷)

This PR is a part of #29

### DIFF
--- a/lib/mobius/actions/message.ex
+++ b/lib/mobius/actions/message.ex
@@ -5,7 +5,10 @@ defmodule Mobius.Actions.Message do
 
   alias Mobius.Models.Message
   alias Mobius.Rest
+  alias Mobius.Rest.Client
   alias Mobius.Services.Bot
+
+  @type message_body :: Rest.Message.message_body()
 
   @doc """
   Send a message in a channel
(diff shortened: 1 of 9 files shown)


lib/mobius/actions/message.ex
```diff
@@ -0,0 +1,52 @@
...
+  @doc """
+  Send a message in a channel
+
+  The params can be:
+    - `content`: The message's content (maximum 2000 chars)
+    - `nonce`: random string or integer to compare with received messages and check its reception
+    - `tts`: true if this message should be read with tts
+    - `embed`: an embed (with type `rich`) with some restrictions
+    - `allowed_mentions`: the mentions allowed in this message (see docs)
+    - `message_reference`: identifies the message being replied to (see docs)
```
Blond11516:
```suggestion
    - `embed`: an embed with type `rich` and some restrictions (see Discord docs)
    - `allowed_mentions`: the mentions allowed in this message (see Discord docs)
    - `message_reference`: identifies the message being replied to (see Discord docs)
```

I know we're in the context of a Discord bot framework but "see docs" inside docs feels a little circular 😛

lib/mobius/actions/message.ex
```diff
@@ -0,0 +1,52 @@
...
+
+  Refer to the documentation for details on the limitations given by Discord.
+  This function is specifically for `content-type: application/json`.
+  See `send_file/3` for `content-type: multipart/form-data`.
+
+  Relevant documentation:
+  https://discord.com/developers/docs/resources/channel#create-message
+  """
+  @spec send_message(keyword, Snowflake.t()) :: Rest.Client.result(Message.t())
+  def send_message(params, channel_id) do
```
Blond11516: Wouldn't it be nicer if the keyword list was the last argument? Are you expecting it to be piped in?

Blond11516 (dismissed):

lib/mobius/actions/message.ex
ZeLarpMaster: I'm indeed expecting it to be piped in so I think this is the most useful for the user. Maybe the params could be a map instead so accessing them for validations is slightly faster and slightly more convenient. This might be relevant since sending messages is probably one of the most common things bots do

Blond11516: Yeah, I agree that if it's not going to be the last argument a map would be better.

Blond11516 (approved):

PR #68, opened 2021-03-08
feat(cogs): Added an api for receiving the context

### LINKED ISSUE #61: Add an API for receiving the context in commands
Right now commands can only receive their arguments in their function, but if we want a way to interact with the user who did the command, we need information about the message which triggered the command.

My current thought is this could be done with a `:context` type of argument in `command/3` which would simply receive the message. This involves the least amount of changes in the current API and is decently clean imo

ZeLarpMaster:
Closes #61

I ended up going with a different api from what was originally intended in the issue because it involved less code changes and was closer to other things like `ExUnit.Case.test/3`

The docs were updated to include plenty of examples and `Mobius.Cog.PingPong` was updated to make use of the context

### DIFF
--- a/lib/mobius/cog.ex
+++ b/lib/mobius/cog.ex
@@ -3,7 +3,7 @@ defmodule Mobius.Cog do
   Defines a cog.
 
   Cogs are small, self-contained services that implement a bot's logic. Cogs
-  have two ways of handling Discord events: `listen/3` and `command/3`.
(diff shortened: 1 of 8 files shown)


Blond11516 (approved):
Great work! It's nice to see that with that and the first actions being implemented we can really start to see the final result forming!

ZeLarpMaster:
I agree! Makes me hyped up for the first cogs we could implement 😄

PR #69, opened 2021-03-10
feat: Add "Create Reaction" action

Blond11516:
Adds the `create_reaction` action.

This PR is a part of https://github.com/ZeLarpMaster/Mobius/issues/29.

### DIFF
--- a/lib/mobius/actions/reaction.ex
+++ b/lib/mobius/actions/reaction.ex
@@ -1,4 +1,8 @@
 defmodule Mobius.Actions.Reaction do
+  @moduledoc """
+  Actions related to Discord reactions such as creating, removing and listing reactions.
+  """
+
   alias Mobius.Rest
   alias Mobius.Services.Bot
   alias Mobius.Models.Snowflake
@@ -9,8 +13,8 @@ defmodule Mobius.Actions.Reaction do
 
   ## Example
 
-      iex> emoji = %Emoji{name: "👌", require_colon: false, managed: false, animated: false, available: true}
-      ...> create_reaction(emoji, channel_id, message_id)
+      iex> emoji = %Mobius.Models.Emoji{name: "👌", require_colons: false, managed: false, animated: false, available: true}
+      ...> Mobius.Actions.Reaction.create_reaction(emoji, channel_id, message_id)
       :ok
 
   ## Documentation
--- a/lib/mobius/actions/reaction.ex
+++ b/lib/mobius/actions/reaction.ex
@@ -1,0 +1,49 @@
+defmodule Mobius.Actions.Reaction do
+  alias Mobius.Rest
+  alias Mobius.Services.Bot
+  alias Mobius.Models.Snowflake
+
+  @doc """
+  Add a reaction to a message with a custom emoji
+
+  ## Example
+
+      iex> create_reaction(channel_id, message_id, "party_parrot", custom_emoji_id)
+      :ok
+
+  ## Documentation
+
+  Relevant documentation:
+  https://discord.com/developers/docs/resources/channel#create-reaction
+  """
+  @spec create_reaction(Snowflake.t(), Snowflake.t(), String.t(), Snowflake.t()) ::
+          Client.empty_result()
+  def create_reaction(channel_id, message_id, custom_emoji_name, custom_emoji_id) do
+    Rest.Reaction.create_reaction(
+      Bot.get_client!(),
+      channel_id,
+      message_id,
+      "#{custom_emoji_name}:#{custom_emoji_id}"
+    )
+  end
+
+  @doc """
+  Add a reaction to a message with a Unicode emoji
+
+  The emoji should be passed as its literal Unicode representation.
+
+  ## Example
+
+      iex> create_reaction(channel_id, message_id, "👌")
+      :ok
+
+  ## Documentation
+
+  Relevant documentation:
+  https://discord.com/developers/docs/resources/channel#create-reaction
+  """
+  @spec create_reaction(Snowflake.t(), Snowflake.t(), String.t()) :: Client.empty_result()
+  def create_reaction(channel_id, message_id, unicode_emoji) do
+    Rest.Reaction.create_reaction(Bot.get_client!(), channel_id, message_id, unicode_emoji)
+  end
+end
--- a/lib/mobius/actions/reaction.ex
+++ b/lib/mobius/actions/reaction.ex
@@ -2,13 +2,15 @@ defmodule Mobius.Actions.Reaction do
   alias Mobius.Rest
   alias Mobius.Services.Bot
   alias Mobius.Models.Snowflake
+  alias Mobius.Models.Emoji
 
   @doc """
-  Add a reaction to a message with a custom emoji
+  Add a reaction to a message
 
   ## Example
 
-      iex> create_reaction(channel_id, message_id, "party_parrot", custom_emoji_id)
+      iex> emoji = %Emoji{name: "👌", require_colon: false, managed: false, animated: false, available: true}
+      ...> create_reaction(emoji, channel_id, message_id)
       :ok
 
   ## Documentation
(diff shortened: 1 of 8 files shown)


lib/mobius/actions/reaction.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Mobius.Actions.Reaction do
+  alias Mobius.Rest
+  alias Mobius.Services.Bot
+  alias Mobius.Models.Snowflake
+
+  @doc """
+  Add a reaction to a message with a custom emoji
```
Blond11516: The doc on both heads is really thin but I feel like the argument names and spec are pretty self-explanatory :shrug:

lib/mobius/actions/reaction.ex
```diff
@@ -0,0 +1,49 @@
...
+      :ok
+
+  ## Documentation
+
+  Relevant documentation:
+  https://discord.com/developers/docs/resources/channel#create-reaction
+  """
+  @spec create_reaction(Snowflake.t(), Snowflake.t(), String.t(), Snowflake.t()) ::
+          Client.empty_result()
+  def create_reaction(channel_id, message_id, custom_emoji_name, custom_emoji_id) do
```
ZeLarpMaster: I have a few suggestions on this API which may or may not be good ideas:
- I think using `Mobius.Models.Emoji.t()` instead of the name and id would be better since it would allow taking emojis from other endpoints or events and send them directly to this call. It would also merge both functions into a single function which makes it easier to take an emoji variable (regardless of whether it's custom or not) and send it to this function (instead of having to do a `if custom emoji, do: create_reaction/4, else: create_reaction/3` everytime). And finally it allows for:
- Putting the emoji as the first argument to make it easier to pipe the emoji from a config or an event into this function. I think you're more likely to grab the emoji from a few other function calls than the channel or message

lib/mobius/rest/client.ex
```diff
@@ -7,6 +7,7 @@ defmodule Mobius.Rest.Client do
   @type errors :: :unauthorized_token | :forbidden | :not_found | :ratelimited | any
   @type error :: {:error, errors()}
   @type result(arg) :: {:ok, arg} | error()
+  @type empty_result :: :ok | error()
```
ZeLarpMaster, line 10: Could replace `check_empty_response`'s return type to this

lib/mobius/rest/reaction.ex
```diff
@@ -0,0 +1,29 @@
...
+      %{},
+      opts: [
+        path_params: [
+          channel_id: channel_id,
+          message_id: message_id,
+          emoji: emoji
+        ]
+      ]
+    )
+    |> Client.parse_response(fn _ -> :ok end)
```
ZeLarpMaster: I think this should be `Client.check_empty_response/1`

test/mobius/actions/reaction_test.exs
```diff
@@ -0,0 +1,63 @@
...
+    test "returns :ok if successful", ctx do
+      :ok = Reaction.create_reaction(ctx.channel_id, ctx.message_id, ctx.emoji)
+    end
+  end
+
+  describe "create_reaction/4" do
+    setup do
+      message_id = random_snowflake()
+      channel_id = random_snowflake()
+      emoji_name = "party-parrot"
```
ZeLarpMaster: funnily enough this is an invalid custom emoji name ("can only contain alphanumeric characters and underscores"), though I guess it doesn't matter at all here

lib/mobius/actions/reaction.ex
ZeLarpMaster: I agree, if more explanation is needed this can be changed in the future

lib/mobius/actions/reaction.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Mobius.Actions.Reaction do
+  alias Mobius.Rest
```
ZeLarpMaster, line 8: I'd still add a `@moduledoc` here even if it's pretty simple just as a header for the docs which will eventually be on hexdocs

ZeLarpMaster (changes requested):
Good work, I'm happy to see you picked a very important one as your first one 😄

My comments are pretty debatable so if you disagree with something, let me know

lib/mobius/rest/reaction.ex
Blond11516: Didn't know this function exists, I'll change it.

lib/mobius/actions/reaction.ex
```diff
@@ -0,0 +1,57 @@
...
+    "#{emoji.name}:#{emoji.id}"
+  end
+
+  defp get_emoji_string(%Emoji{managed: false, name: nil}) do
+    {:error, "Built-in emojis require a name"}
+  end
+
+  defp get_emoji_string(%Emoji{managed: false} = emoji) do
+    emoji.name
+  end
```
ZeLarpMaster: I like the idea of this function, it might be useful in `Mobius.Models.Emoji` since I'd guess it will be useful elsewhere.

There's a little problem though, `managed` actually means managed by a 3rd party. It's used for things like emotes from Twitch on servers binded to Twitch accounts. It's a little weird so if this doesn't make much sense, do tell me.

Instead I think this function can simply be (though it could be formatted differently)
```elixir
defp get_emoji_string(%Emoji{id: nil, name: name}), do: name
defp get_emoji_string(%Emoji{id: id, name: name}), do: "#{name}:#{id}"
```
This would match what other libs do (for example [discord.py](https://github.com/Rapptz/discord.py/blob/7836046621d6b670730217ad4de321ef3ecd5cff/discord/partial_emoji.py#L134))

ZeLarpMaster (changes requested):
Almost there, just a little thing which is unclear in the docs. After that I think it's gonna be good to go 👍

lib/mobius/actions/reaction.ex
Blond11516: Yeah, I really wasn't sure if there was a way to tell whether an emoji is custom or not since the doc apparently makes no mention of this whatsoever. Not sure where I got the idea that "managed" means "custom" though :thinking:

lib/mobius/actions/reaction.ex
```diff
@@ -0,0 +1,39 @@
...
+      ...> Mobius.Actions.Reaction.create_reaction(emoji, channel_id, message_id)
+      :ok
+
+  ## Documentation
+
+  Relevant documentation:
+  https://discord.com/developers/docs/resources/channel#create-reaction
+  """
+  @spec create_reaction(Emoji.t(), Snowflake.t(), Snowflake.t()) ::
+          Client.empty_result() | {:error, String.t()}
```
ZeLarpMaster: `{:error, String.t()}` is included in `Client.empty_result()`

lib/mobius/actions/reaction.ex
```diff
@@ -0,0 +1,39 @@
...
+  alias Mobius.Models.Snowflake
+  alias Mobius.Rest
+  alias Mobius.Services.Bot
+
+  @doc """
+  Add a reaction to a message
+
+  ## Example
+
+      iex> emoji = %Mobius.Models.Emoji{name: "👌", require_colons: false, managed: false, animated: false, available: true}
```
ZeLarpMaster: I don't think you need to specify all the fields in here, it should be enough to only specify `name:`

lib/mobius/models/emoji.ex
```diff
@@ -49,5 +49,30 @@ defmodule Mobius.Models.Emoji do
...
+  For built-in/Unicode emojis, this is the name, which will be the emoji itself.
+  For custom emojis, this is a string with the format "emoji_name:emoji_id".
+
+  ## Example
+
+      iex> emoji = %Mobius.Models.Emoji{name: "👌", require_colons: false, managed: false, animated: false, available: true}
+      ...> Mobius.Models.Emoji.get_identifier(emoji)
+      "👌"
+
+      iex> emoji = %Mobius.Models.Emoji{id: 123456, name: "👌", require_colons: false, managed: false, animated: false, available: true}
```
ZeLarpMaster: Same thing here

ZeLarpMaster (dismissed):
I'm fine with you merging this as is, but I'd also prefer if you made these tiny changes 😄

ZeLarpMaster (approved):
Good to go!

PR #70, opened 2021-03-10
feat(models): Added Guild

ZeLarpMaster:
Same thing as #37 but for Guild


Blond11516 (approved):
I'll admit that my review is pretty shallow, but I didn't spot any missing fields, so this should be good to go.

PR #73, opened 2021-03-17
fix: Fixing a bunch of TODOs

ZeLarpMaster:
I searched for `TODO` in the code and fixed what was fixable.

If anything is unclear on why it changed, refer to the commit messages. If it's still unclear feel free to ask in here.


lib/mobius/cog.ex
```diff
@@ -228,7 +228,12 @@ defmodule Mobius.Cog do
   myBot: Invalid type for argument "num2". Expected "integer", got "hello".
   """
   defmacro command(command_name, context, args, do: block) do
-    # TODO: assert that the command name contains no whitespace
+    if not Regex.match?(~r/^[a-z0-9_]+$/, command_name) do
```
Blond11516:
```suggestion
    if not Regex.match?(~r/^[[:lower:][:digit:]_]+$/, command_name) do
```

I don't have a strong opinion on readability here, but [character classes](https://en.wikipedia.org/wiki/Regular_expression#Character_classes) are more versatile.

lib/mobius/core/shard_list.ex
```diff
@@ -16,13 +16,14 @@ defmodule Mobius.Core.ShardList do
...
-  def are_all_shards_ready?(table) do
+  @doc "True if any shards have state `:ready`"
+  @spec is_any_shard_ready?(:ets.tab()) :: boolean
+  def is_any_shard_ready?(table) do
     # The match spec was generated with the following line in iex:
-    # :ets.fun2ms(fn {_, state} when state != :ready -> state end)
-    spec = [{{:_, :"$1"}, [{:"/=", :"$1", :ready}], [:"$1"]}]
-    :ets.select(table, spec) == []
+    # :ets.fun2ms(fn {_, state} when state == :ready -> state end)
+    spec = [{{:_, :"$1"}, [{:==, :"$1", :ready}], [:"$1"]}]
```
Blond11516: What would you think of using [ex2ms](https://github.com/ericmj/ex2ms) instead so we can get rid of the ugly match specs and comments?

Blond11516 (dismissed):

Blond11516 (approved):

PR #77, opened 2021-03-30
feat: Added the global command prefix

### LINKED ISSUE #75: Configure a prefix for commands
Bots usually have a prefix (ie.: `!`, `&`, `!?`, etc.) which must be put before commands. This helps to make sure users are talking to the bot they want when multiple bots are on the same server since they'd all have different prefixes. The prefix should be configurable.

ZeLarpMaster:
Closes #75 

A later PR could implement guild-specific prefixes, but that's not a requirement for v1

### DIFF
--- a/test/mobius/core/command_test.exs
+++ b/test/mobius/core/command_test.exs
@@ -41,7 +41,7 @@ defmodule Mobius.Core.CommandTest do
     end
   end
 
-  describe "execute_command/2" do
+  describe "execute_command/3" do
     setup do
       command = %Command{
         name: "hello",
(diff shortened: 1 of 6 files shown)


test/mobius/core/command_test.exs
```diff
@@ -58,27 +58,35 @@ defmodule Mobius.Core.CommandTest do
     end
 
     test "should return an error when no command matches the message" do
-      assert Command.execute_command([], message("hello")) == :not_a_command
+      assert Command.execute_command([], "!", message("hello")) == :not_a_command
```
Blond11516, line 61: Should we add a small helper for that? It could remove some boilerplate, but I'm not if it would make the tests less clear...

Blond11516 (approved):

test/mobius/core/command_test.exs
ZeLarpMaster: Considering we'd still need to pass in `[]` and `"hello"`, it would look like `assert execute_command([], "hello") == :not_a_command` which is very similar in terms of reading, but also hides the function we're testing here behind another function which makes it harder to follow. So I'm not really a fan about that in this case

PR #78, opened 2021-04-07
feat: Added Help cog

### LINKED ISSUE #76: Write a help command
The help command has 3 different usages:
- `[p]help` lists all cogs and all of their commands with a short description of each
- `[p]help [cog]` lists all commands in the cog with a short description of the commands and a description of the cog
- `[p]help [command]` gives a long description of the command with the full spec of how to use it (ie.: `[p]command [arg1] [arg2]`)

ZeLarpMaster:
Added the `Help` cog which supports:
- [x] `[p]help`
- [x] `[p]help <cog>`
- [x] `[p]help <command>`

The cog is autoloaded by default.

Closes #76

### DIFF
--- a/lib/mobius/cog_utils.ex
+++ b/lib/mobius/cog_utils.ex
@@ -3,7 +3,9 @@ defmodule Mobius.CogUtils do
   A module of utilities for cogs such as formatting tables
   """
 
-  @type categories :: [{String.t(), [{String.t(), String.t()}]}]
+  @type category_entry :: {String.t(), String.t()}
+  @type category :: {String.t(), [category_entry()]}
+  @type categories :: [category()]
 
   @doc ~S"""
   Formats a list of tuples such that the spacing aligns the 2nd element of the inner tuple
@@ -12,14 +14,26 @@ defmodule Mobius.CogUtils do
 
       iex> data = [{"Cat1", [{"abc", "description"}, {"defghi", "something"}]}]
       iex> format_categories_list(data)
-      "Cat1:\n  abc       description\n  defghi    something"
+      ~S"
+      Cat1:
+        abc       description
+        defghi    something
+      " |> String.trim()
 
       iex> data = [{"Cat1", [{"abc", "description"}]}, {"Cat2", [{"defghi", "something"}]}]
       iex> format_categories_list(data)
-      "Cat1:\n  abc       description\nCat2:\n  defghi    something"
+      ~S"
+      Cat1:
+        abc       description
+      Cat2:
+        defghi    something
+      " |> String.trim()
 
       iex> format_categories_list([{"Cat1", []}], "empty category")
-      "Cat1:\n  empty category"
+      ~S"
+      Cat1:
+        empty category
+      " |> String.trim()
 
       iex> format_categories_list([])
       ""
--- a/lib/mobius/cog_utils.ex
+++ b/lib/mobius/cog_utils.ex
@@ -1,0 +1,50 @@
+defmodule Mobius.CogUtils do
+  @moduledoc """
+  A module of utilities for cogs such as formatting tables
+  """
+
+  @type categories :: [{String.t(), [{String.t(), String.t()}]}]
+
+  @doc ~S"""
+  Formats a list of tuples such that the spacing aligns the 2nd element of the inner tuple
+
+  ## Example
+
+      iex> data = [{"Cat1", [{"abc", "description"}, {"defghi", "something"}]}]
+      iex> format_categories_list(data)
+      "Cat1:\n  abc       description\n  defghi    something"
+
+      iex> data = [{"Cat1", [{"abc", "description"}]}, {"Cat2", [{"defghi", "something"}]}]
+      iex> format_categories_list(data)
+      "Cat1:\n  abc       description\nCat2:\n  defghi    something"
+
+      iex> format_categories_list([{"Cat1", []}], "empty category")
+      "Cat1:\n  empty category"
+
+      iex> format_categories_list([])
+      ""
+  """
+  @spec format_categories_list(categories(), String.t()) :: String.t()
+  def format_categories_list(list, empty_category \\ "nothing here") do
+    column_width =
+      list
+      |> Enum.flat_map(fn {_category, rows} -> rows end)
+      |> Enum.map(fn {name, _description} -> String.length(name) end)
+      |> Enum.max(&>=/2, fn -> 0 end)
+
+    list
+    |> Enum.map(fn
+      {category, []} -> "#{category}:\n  #{empty_category}"
+      {category, rows} -> "#{category}:\n#{format_rows(rows, column_width)}"
+    end)
+    |> Enum.join("\n")
+  end
+
+  defp format_rows(rows, width) do
+    rows
+    |> Enum.map(fn {name, description} ->
+      "  " <> String.pad_trailing(name, width) <> "    " <> description
+    end)
+    |> Enum.join("\n")
+  end
+end
--- a/lib/mobius/cogs/help.ex
+++ b/lib/mobius/cogs/help.ex
@@ -5,10 +5,20 @@ defmodule Mobius.Cogs.Help do
 
   import Mobius.Actions.Message
 
+  alias Mobius.Core.Command
   # Unsafe to use for 3rd party cogs
   alias Mobius.Services.CogLoader
 
-  @doc "Wow"
+  @header """
+  Type `[p]help Cog` for help about a specific cog
+  Type `[p]help command` for help about a specific command
+  """
+
+  @footer """
+  Made with Mobius, a general purpose bot written in Elixir
+  """
+
+  @doc "This command"
   command "help", context do
     cogs = CogLoader.list_cogs()
 
@@ -17,19 +27,24 @@ defmodule Mobius.Cogs.Help do
       |> Enum.map(&format_cog/1)
       |> Enum.join("\n")
 
-    send_message(%{content: "```#{cogs_list}```"}, context.channel_id)
+    send_message(%{content: "#{@header}```#{cogs_list}```#{@footer}"}, context.channel_id)
   end
 
   defp format_cog(%Mobius.Cog{name: name, commands: commands}) do
     commands_list =
       commands
+      |> Enum.filter(fn %Command{description: description} -> description != false end)
       |> Enum.map(&format_command/1)
       |> Enum.join("\n")
 
-    "#{name}:\n#{commands_list}"
+    if commands_list == "" do
+      "#{name}:\n    has no commands"
+    else
+      "#{name}:\n#{commands_list}"
+    end
   end
 
-  defp format_command(%Mobius.Core.Command{name: name, description: description}) do
+  defp format_command(%Command{name: name, description: description}) do
     "    #{name}        #{description}"
   end
 end
--- a/lib/mobius/cogs/help.ex
+++ b/lib/mobius/cogs/help.ex
@@ -19,15 +19,19 @@ defmodule Mobius.Cogs.Help do
   """
 
   @doc "This command"
-  command "help", context do
+  command "help", context, part: :string do
     cogs = CogLoader.list_cogs()
 
+    send_message(%{content: format_cogs(cogs)}, context.channel_id)
+  end
+
+  defp format_cogs(cogs) do
     cogs_list =
       cogs
       |> Enum.map(&format_cog/1)
       |> Enum.join("\n")
 
-    send_message(%{content: "#{@header}```#{cogs_list}```#{@footer}"}, context.channel_id)
+    "#{@header}```#{cogs_list}```#{@footer}"
   end
 
   defp format_cog(%Mobius.Cog{name: name, commands: commands}) do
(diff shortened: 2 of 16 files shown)


lib/mobius/cog_utils.ex
```diff
@@ -0,0 +1,50 @@
...
+  @type categories :: [{String.t(), [{String.t(), String.t()}]}]
+
+  @doc ~S"""
+  Formats a list of tuples such that the spacing aligns the 2nd element of the inner tuple
+
+  ## Example
+
+      iex> data = [{"Cat1", [{"abc", "description"}, {"defghi", "something"}]}]
+      iex> format_categories_list(data)
+      "Cat1:\n  abc       description\n  defghi    something"
```
Blond11516:
```suggestion
      ~S"
      Cat1:
        abc       description
        defghi    something
      "
      |> String.trim()
```
What do you think of that? It's definitely more verbose, but I think it's much more clear as to how the input gets formatted exactly.

lib/mobius/cog_utils.ex
```diff
@@ -0,0 +1,50 @@
...
+
+      iex> data = [{"Cat1", [{"abc", "description"}]}, {"Cat2", [{"defghi", "something"}]}]
+      iex> format_categories_list(data)
+      "Cat1:\n  abc       description\nCat2:\n  defghi    something"
```
Blond11516:
```suggestion
      ~S"
      Cat1:
        abc       description
      Cat2:
        defghi    something
      "
      |> String.trim()
```

lib/mobius/cog_utils.ex
```diff
@@ -0,0 +1,50 @@
...
+
+      iex> format_categories_list([{"Cat1", []}], "empty category")
+      "Cat1:\n  empty category"
```
Blond11516:
```suggestion
      ~S"
      Cat1:
        empty category
      "
      |> String.trim()
```

lib/mobius/cog_utils.ex
```diff
@@ -0,0 +1,50 @@
+defmodule Mobius.CogUtils do
+  @moduledoc """
+  A module of utilities for cogs such as formatting tables
+  """
+
+  @type categories :: [{String.t(), [{String.t(), String.t()}]}]
```
Blond11516: That's a little dense. I think it would be worth it to extract parts of this to named types. What would you think of
```elixir
@type category_entry :: {String.t(), String.t()}
@type category :: {String.t(), [category_entry()]}
@type categories :: [category()] # Might not event be necessary, could easily be moved to the function spec
```

lib/mobius/cogs/help.ex
```diff
@@ -0,0 +1,123 @@
...
+  The name is case sensitive to distinguish between cog names and command names.
+  For example `[p]help Help` shows the cog, but `[p]help help` shows the command.
+  """
+  command "help", cog_or_command_name: :string do
+    cogs = CogLoader.list_cogs()
+
+    cog_or_command_name
+    |> try_cog(cogs)
+    |> try_command(cog_or_command_name, cogs)
+    |> reply()
```
Blond11516: What would you think of an "inverse with" like we did the other day?
```elixir
with nil <- try_cog(cog_or_command_name, cogs),
     nil <- try_command(cog_or_command_name, cogs) do
  {:reply, %{content: @not_found}}
else
  content -> {:reply, %{content: content, allowed_mentions: %{parse: []}}}
end
```
This probably is the opposite of idiomatic, but it feels a little weird to me that `try_command` has to worry about the fact that there might already be a result available...

lib/mobius/cogs/help.ex
```diff
@@ -0,0 +1,123 @@
...
+    cogs_list =
+      cogs
+      |> Enum.filter(fn %Cog{description: description} -> description != false end)
+      |> Enum.map(fn %Cog{name: name} = cog -> {name, list_cog_commands(cog)} end)
+      |> CogUtils.format_categories_list("has no commands")
+
+    "```#{cogs_list}```#{@help_footer}"
+  end
+
+  defp list_cog_commands(%Cog{commands: commands}) do
```
Blond11516, line 100: Shouldn't that be a concern of the Cog module?

lib/mobius/cogs/help.ex
```diff
@@ -0,0 +1,123 @@
...
+  defp try_command(content, _part, _cogs) when is_binary(content), do: content
+
+  defp try_command(nil, part, cogs) do
+    case find_command(part, cogs) do
+      arities when is_map(arities) -> format_specific_command(arities)
+      _ -> nil
+    end
+  end
+
+  defp format_specific_cog(%Cog{} = cog) do
```
Blond11516: For the formatting functions, what would you think of "format_x"/"format_x_list" instead of "format_specific_x"/"format_xs"?

lib/mobius/cogs/help.ex
```diff
@@ -0,0 +1,123 @@
...
+    commands = list_cog_commands(cog)
+    command_list = CogUtils.format_categories_list([{"Commands", commands}])
+
+    "#{cog.description}\n```#{command_list}```#{@cog_footer}"
```
Blond11516:
```suggestion
    formatted_commands = CogUtils.format_categories_list([{"Commands", commands}])

    "#{cog.description}\n```#{formatted_commands}```#{@cog_footer}"
```
I feel like "list" could be misleading here

lib/mobius/cogs/help.ex
Blond11516: Right now it's pretty hard to get a sense of what the formatting functions' output is actually supposed to look like. I think it would be worth it to extract them to a `HelpFormatter` module and add doc(test)s to them, just to give concrete examples. I wouldn't add any additional unit tests though.

test/support/cog_test_utils.ex
```diff
@@ -38,19 +39,16 @@ defmodule Mobius.CogTestUtils do
...
-  `%{tts: false, content: "Content"}`
-  """
-  @spec assert_message_sent(Message.message_body()) :: :ok
-  def assert_message_sent(message) do
-    json_message = Jason.encode!(message)
-    %Tesla.Env{url: url} = assert_receive %Tesla.Env{body: ^json_message, method: :post}
-    # This only works if the channel_id used to send the message was `nil`
-    assert String.ends_with?(url, "/channels/:channel_id/messages")
-    :ok
+  @doc "Asserts that a message was sent returns the body of the request"
```
Blond11516:
```suggestion
  @doc "Asserts that a message was sent and returns the body of the request"
```
The end isn't quite true, the body isn't returned, but merely matched against the expectation. I think that just saying "and matches the body of the request against the expectation" might not be clear enough even though it's technically more accurate. What do you think?

Blond11516 (changes requested):

lib/mobius/cog_utils.ex
ZeLarpMaster: It messes with the spacing on the 2nd and 3rd lines though. I don't think there's a quick and easy way to make it work. Also, is the expected value executed? I thought it had to be matching what's printed or something like that 🤔 I'll try to make it better, but I'm a bit skeptical

lib/mobius/cog_utils.ex
ZeLarpMaster: Oh you're absolutely right

lib/mobius/cogs/help.ex
ZeLarpMaster: Yeah it's probably opposite of idiomatic, but I don't really mind. imo it still reads better than the other options

lib/mobius/cogs/help.ex
ZeLarpMaster: I don't think so, the function actually goes in and generates tuples of `{command_name, description}` which I think is very specific to what's going on here. If there's a need for something very similar elsewhere, maybe we can put it together, but right now I really feel like this is specific enough to warrant being here

lib/mobius/cogs/help.ex
ZeLarpMaster: Sure, I agree it would be more readable 👍

ZeLarpMaster: Sounds good! It will also help with extracting it into a behavior for a future where we wanna output the help into an Embed instead of directly in the message.

test/support/cog_test_utils.ex
ZeLarpMaster: Yeah I changed what this function does like 4 times while doing this PR and probably forgot to update the doc at some point 😅

It does however return the body of the request because the `assert unquote(x) = body` is the last thing in the quote which means `body = assert_message_sent(%{})` does in fact have the body of the request in the variable `body`. It's not really the expected way to use this macro though, I agree.

I think something like "Asserts and matches on the body of a message send request" with an example might be good enough? I don't really see how to be clear and concise here so I'd embrace it by providing examples instead.

lib/mobius/cog_utils.ex
Blond11516: Did you test it? It works as expected for me.

image
image
image

test/support/cog_test_utils.ex
Blond11516: You're right, I forgot `assert` actually returns the result! And I agree, examples are probably the way to go here.

lib/mobius/cogs/help.ex
```diff
@@ -0,0 +1,123 @@
...
+
+  defp list_cog_commands(%Cog{commands: commands}) do
+    commands
+    |> Enum.sort_by(&elem(&1, 0))
+    |> Enum.map(fn {name, arities} -> {name, find_command_description(arities)} end)
+  end
+
+  defp find_command_description(arities) do
+    arities
+    |> Enum.min_by(&elem(&1, 0))
```
Blond11516: `list_cog_commands` and `find_command_description` use `elem/2` a total of 3 times. This means that unless I have a very good understanding of how the command struct is built, it's pretty hard to know what's going on here. I think changing the `elem` calls to anonymous functions with explicit pattern matching would go a long way to make this clearer.

lib/mobius/cogs/help.ex
Blond11516: There is an element of "tell don't ask" here though. This came to mind while writing my comment about `elem`, but should the help cog really be bothered so much with the internals of a cog? Maybe the function needs to be reworked a bit (ideally I'd see it returning `[%Command.t()]`) because I agree the exact use case here is pretty specific to the help command.

lib/mobius/cog_utils.ex
ZeLarpMaster: I'm so confused by how this works. I'm expecting 6 spaces before "Cat1" in your screenshots and I don't get why that's not the case. I agree it's much better, I just didn't know it worked like that. Thank you!

lib/mobius/cogs/help.ex
ZeLarpMaster: If you have a specific suggestion, I'm open to changing how it works, but I fail to see how we could make `Cog` return anything different and still useful here. The help command needs to list the commands of each cog together and then the description of each command (and sort all of that alphabetically). If we return `[%Command.t()]` from a function in Cog, we're doing work to flatten the structure, and then we're doing the exact opposite work to group the commands of each cog and then group the clauses of each command.

Maybe I'm just getting your comment wrong, in which case I apologize.

lib/mobius/cog_utils.ex
Blond11516: I guess the doctest algorithm accounts for that and "aligns" the sigil with the `iex>` prompt 🤷

lib/mobius/cogs/help.ex
Blond11516: When you say we'd be doing work to flatten the structure and then regroup it are you thinking that "the function that would live inside the Cog module" (copyright pending) would return the list of _all_ commands? Cuz that's not what I had in mind.

I was just thinking that we could move this whole function to `Cog` in either one of two ways:
1. Move it as-is and maybe rename it to something like `get_sorted_command_descriptions`
2. Split and move it to something like `get_sorted_commands` and `get_command_description`. `Help` would then do something like `cog |> Cog.get_sorted_commands() |> Enum.map(&Cog.get_command_description/1)`

(Not sure how the second options would actually end up looking like, so the first might be better)

Again, for me this really boils down to "tell don't ask". The help command is doing a lot of work with the _internal_ representation of the cogs and commands, and I don't think it should have to.

Blond11516: Maybe we need to talk about this over voice. Let me know if you want to do that!

lib/mobius/core/command.ex
```diff
@@ -57,6 +57,19 @@ defmodule Mobius.Core.Command do
     |> Map.new(fn {name, commands} -> {name, Enum.group_by(commands, &arg_count/1)} end)
   end
 
+  @doc """
+  Returns a description for a command based on its `t:command_arities/0`
+
+  It returns the description of the first clause which has a description out of the clauses
+  with the lowest arity
```
Blond11516:
```suggestion
  with the lowest arity.
```

Blond11516 (dismissed):

Blond11516 (approved):

PR #79, opened 2021-04-09
feat: Supporting overloaded commands

ZeLarpMaster:
This PR is the first step towards supporting everything needed for #76 (cog docs, command docs, and command overloading)

This PR adds support for command overloading in a naive way. A future PR will refactor this implementation to make it both more efficient and more readable. Since that refactoring will touch multiple modules, I preferred splitting the PR into one which adds the feature and another which refactors it.

Note: command overloading means the same command with a different arity and/or different types on the arguments. This implementation supports both different arities and different types.

### DIFF
--- a/lib/mobius/core/command.ex
+++ b/lib/mobius/core/command.ex
@@ -36,32 +36,44 @@ defmodule Mobius.Core.Command do
 
   @spec execute_command([t()], String.t(), Message.t()) :: handle_message_result()
   def execute_command(commands, prefix, %Message{content: content} = message) do
-    commands
-    |> Enum.find(fn %__MODULE__{} = command ->
-      String.starts_with?(content, prefix <> command.name)
-    end)
-    |> case do
-      nil ->
-        :not_a_command
+    arg_values = split_arguments(content)
 
-      %__MODULE__{} = command ->
-        arg_values = split_arguments(content)
-
-        with :ok <- validate_arg_count(command, arg_values),
-             {:ok, values} <- parse_arg_values(command, arg_values) do
-          {:ok, apply(command.handler, [message | values])}
+    Enum.reduce_while(commands, :not_a_command, fn %__MODULE__{} = command, acc ->
+      new =
+        cond do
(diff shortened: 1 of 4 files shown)


lib/mobius/core/command.ex
```diff
@@ -36,32 +36,44 @@ defmodule Mobius.Core.Command do
...
+      {{:ok, _} = value, _} -> {:halt, value}
+      {_, {:ok, _} = value} -> {:halt, value}
+      {{:invalid_args, _} = value, _} -> {:cont, value}
+      {_, {:invalid_args, _} = value} -> {:cont, value}
+      {{:too_few_args, _, _} = value, _} -> {:cont, value}
+      {_, {:too_few_args, _, _} = value} -> {:cont, value}
+      {:not_a_command, _} -> {:cont, :not_a_command}
+      {_, :not_a_command} -> {:cont, :not_a_command}
+    end
+  end
```
Blond11516, line 72: That is _extremely_ dense. I don't know what your exact plans are for the refactor, but I think changing the data structure for commands would be a huge win here (I'm thinking grouping commands by name under a map at least, and then maybe grouping them again by arg count).

test/mobius/cog_test.exs
```diff
@@ -55,34 +55,33 @@ defmodule Mobius.CogTest do
       assert_receive "hello"
     end
 
-    test "should parse integer arguments" do
-      send_command_payload("add 1 2")
-
-      assert_receive 3
+    test "should call the clause with matching types" do
+      send_command_payload("reply 5")
+      assert_receive 15
```
Blond11516: Shouldn't `reply 5` send 5? I know this is what the cog stub does, but maybe supporting integers and strings for `add` would be a clearer way to test this?

Blond11516 (dismissed):

lib/mobius/core/command.ex
ZeLarpMaster: That's exactly what I had in mind too :)

Blond11516 (approved):

PR #82, opened 2021-04-10
feat(actions): Implement get channel action

Blond11516:
This PR is a part of #29.

### DIFF
--- a/lib/mobius/actions/channel.ex
+++ b/lib/mobius/actions/channel.ex
@@ -4,6 +4,7 @@ defmodule Mobius.Actions.Channel do
   """
 
   alias Mobius.Models.Channel
+  alias Mobius.Models.Snowflake
   alias Mobius.Rest
   alias Mobius.Rest.Client
   alias Mobius.Services.Bot
@@ -23,8 +24,6 @@ defmodule Mobius.Actions.Channel do
   """
   @spec get(Snowflake.t()) :: Client.result(Channel.t())
   def get(channel_id) do
-    # TODO: Make sure the channel exists (requires a cache)
-
     if Bot.ready?() do
       Rest.Channel.get(Bot.get_client!(), channel_id)
     else
--- a/lib/mobius/actions/channel.ex
+++ b/lib/mobius/actions/channel.ex
@@ -22,8 +22,8 @@ defmodule Mobius.Actions.Channel do
   Relevant documentation:
   https://discord.com/developers/docs/resources/channel#get-channel
   """
-  @spec get(Snowflake.t()) :: Client.result(Channel.t())
-  def get(channel_id) do
-    Rest.Channel.get(Bot.get_client!(), channel_id)
+  @spec get_channel(Snowflake.t()) :: Client.result(Channel.t())
+  def get_channel(channel_id) do
+    Rest.Channel.get_channel(Bot.get_client!(), channel_id)
   end
 end
--- a/lib/mobius/actions/channel.ex
+++ b/lib/mobius/actions/channel.ex
@@ -24,10 +24,6 @@ defmodule Mobius.Actions.Channel do
   """
   @spec get(Snowflake.t()) :: Client.result(Channel.t())
   def get(channel_id) do
-    if Bot.ready?() do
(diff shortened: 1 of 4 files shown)


lib/mobius/actions/channel.ex
```diff
@@ -0,0 +1,34 @@
...
+  ## Documentation
+
+  Relevant documentation:
+  https://discord.com/developers/docs/resources/channel#get-channel
+  """
+  @spec get(Snowflake.t()) :: Client.result(Channel.t())
+  def get(channel_id) do
+    # TODO: Make sure the channel exists (requires a cache)
+
+    if Bot.ready?() do
```
Blond11516: Are we gonna want to do this check for all calls? If so I think we should probably create a wrapper around Tesla and do the check there.

ZeLarpMaster: Only create message has something about that in the docs:
image

I really don't know whether we can actually do stuff like `get_channel` without being connected to the socket though

Blond11516: Hmm, that's interesting :thinking: I'll try to look into that a bit.

lib/mobius/actions/channel.ex
```diff
@@ -0,0 +1,34 @@
...
+      {:ok, %Mobius.Models.Channel{} = channel}
+
+  ## Documentation
+
+  Relevant documentation:
+  https://discord.com/developers/docs/resources/channel#get-channel
+  """
+  @spec get(Snowflake.t()) :: Client.result(Channel.t())
+  def get(channel_id) do
+    # TODO: Make sure the channel exists (requires a cache)
```
ZeLarpMaster: We don't actually want to check the cache for this, we're giving users access to the rest api's get so they can manually fetch by api if it isn't in cache or for whatever other reason

lib/mobius/actions/channel.ex
```diff
@@ -0,0 +1,34 @@
...
+
+      iex> get(132456789)
+      {:ok, %Mobius.Models.Channel{} = channel}
+
+  ## Documentation
+
+  Relevant documentation:
+  https://discord.com/developers/docs/resources/channel#get-channel
+  """
+  @spec get(Snowflake.t()) :: Client.result(Channel.t())
```
ZeLarpMaster: I think you're missing the `alias` for Snowflake 🤔

lib/mobius/actions/channel.ex
Blond11516: Well, I commented out the `identify` call and both `send_message` and `Channel.get` still work fine...

Blond11516: Actually I commented out both `identify` and the gateway socket creation itself and both actions still work fine...

ZeLarpMaster: That's surprising, makes me wonder what the warning is actually about 🤔 It does say "at least once" so I wonder on a new bot account if it would still work without ever connecting to the gateway?

If you can send a message without ever being connected, I think we'd be able to remove that check from both places.

ZeLarpMaster: ~~I tried with my token and a new account token and both give me a 401: 
image
It does that on both Get Channel and Create Message~~

NVM all of that, I was sending the wrong auth header :(

ZeLarpMaster: Ok so after trying again but properly, I get this when I try to send a message without ever connecting to the gateway
image
But I get the correct (200) response from Get Channel

ZeLarpMaster: I tried again with another token of a bot I know hasn't been online once in the last few months and I was able to do both Create Message and Get Channel without connecting to the gateway. So I think Discord really means "at least once" as in "at least once ever".

My conclusion from that is we can keep the "ready" check for sending messages to make sure the bot has connected at least once before sending messages, but we don't need the check for anything else (unless the warning is there on another call somewhere)

Blond11516: Wow good catch, hadn't thought of it like that at all! I wonder why they'd add such a restriction though :thinking:

ZeLarpMaster (approved):

PR #83, opened 2021-04-11
refactor: Command execution

### LINKED ISSUE #80: feat: Support spaces in the prefix
Spaces are currently not supported in command prefixes because `Command` simply skips all characters before the first space (see `split_arguments/1`). The best way of confirming this works through time would be to change the default prefix for tests to something with a space (`sudo` followed by a space is a good example).

ZeLarpMaster:
This PR is the followup to #79 and is the refactor talked about in there.

While doing the refactor, I found an easy way to fix #80 so that's also done :) It would have been inconvenient to not fix it at the same time. As such I changed the default prefix for tests to be "sudo " which means a lot of existing tests will be testing that prefixes with spaces work.

The pingpong cog has also been fixed. It was forgotten after #72

Note: You may notice the error messages in Cog are now partly less informative. I haven't found an easy way to do the same error messages now that multiple clauses are there. I also didn't want to spend too much effort on that so I've put in as much information as available where the errors are detected and put that in the error tuple. At the end of the day I think it's fine anyway since later on the error won't be logged, it will instead make the bot reply with the command's help message which should list all the clauses. This is an overall much better error message for end users too (especially compared to not replying at all)

### DIFF
--- a/lib/mobius/cog.ex
+++ b/lib/mobius/cog.ex
@@ -83,17 +83,15 @@ defmodule Mobius.Cog do
 
       @computed_commands Command.preprocess_commands(@commands)
 
-      defp __commands__, do: @computed_commands
-
       listen :message_create, message do
-        case Command.execute_command(__commands__(), Bot.get_global_prefix!(), message) do
+        case Command.execute_command(@computed_commands, Bot.get_global_prefix!(), message) do
           {:ok, _} ->
             :ok
 
           {:too_few_args, arities, received} ->
             Logger.info(
               "Wrong number of arguments. Expected one of #{
-                Enum.join(Enum.map(arities, &Integer.to_string/1), ", ")
+                arities |> Enum.map(&Integer.to_string/1) |> Enum.join(", ")
               } arguments, got #{received}."
             )
 
--- a/lib/mobius/cog.ex
+++ b/lib/mobius/cog.ex
@@ -81,28 +81,26 @@ defmodule Mobius.Cog do
       alias Mobius.Actions.Events
       alias Mobius.Services.Bot
 
-      @computed_commands Enum.reverse(@commands)
+      @computed_commands Command.preprocess_commands(@commands)
+
+      defp __commands__, do: @computed_commands
 
       listen :message_create, message do
-        case Command.execute_command(@computed_commands, Bot.get_global_prefix!(), message) do
+        case Command.execute_command(__commands__(), Bot.get_global_prefix!(), message) do
           {:ok, _} ->
             :ok
 
-          {:too_few_args, command, received} ->
+          {:too_few_args, arities, received} ->
             Logger.info(
-              "Wrong number of arguments for command \"#{command.name}\". Expected #{
-                Command.arg_count(command)
+              "Wrong number of arguments. Expected one of #{
+                Enum.join(Enum.map(arities, &Integer.to_string/1), ", ")
               } arguments, got #{received}."
             )
 
-          {:invalid_args, errors} ->
-            Enum.each(errors, fn {{arg_name, arg_type}, value} ->
-              Logger.info(
-                ~s(Invalid type for argument "#{arg_name}". Expected "#{Atom.to_string(arg_type)}", got "#{
-                  value
-                }".)
-              )
-            end)
+          {:invalid_args, [clause | _clauses]} ->
+            Logger.info(
+              ~s(Type mismatch for the command "#{clause.name}" with #{Command.arg_count(clause)} arguments)
+            )
 
           :not_a_command ->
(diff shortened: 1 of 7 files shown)


lib/mobius/cog.ex
```diff
@@ -81,28 +81,26 @@ defmodule Mobius.Cog do
       alias Mobius.Actions.Events
       alias Mobius.Services.Bot
 
-      @computed_commands Enum.reverse(@commands)
+      @computed_commands Command.preprocess_commands(@commands)
+
+      defp __commands__, do: @computed_commands
```
Blond11516: What's the advantage of using that instead of the module attribute?

lib/mobius/cog.ex
```diff
@@ -81,28 +81,26 @@ defmodule Mobius.Cog do
...
           {:ok, _} ->
             :ok
 
-          {:too_few_args, command, received} ->
+          {:too_few_args, arities, received} ->
             Logger.info(
-              "Wrong number of arguments for command \"#{command.name}\". Expected #{
-                Command.arg_count(command)
+              "Wrong number of arguments. Expected one of #{
+                Enum.join(Enum.map(arities, &Integer.to_string/1), ", ")
```
Blond11516:
```suggestion
                arities |> Enum.map(&Integer.to_string/1) |> Enum.join(", ")
```

Don't know how you feel about single line pipes but I think this reads much better.

lib/mobius/core/command.ex
```diff
@@ -34,49 +40,44 @@ defmodule Mobius.Core.Command do
...
-
-      decide_by_priority(acc, new)
-    end)
+    with {:ok, content} <- match_prefix(prefix, content),
+         {:ok, name, arg_values} <- split_arguments(content),
+         {:ok, groups} <- get_command(commands, name),
+         {:ok, clauses} <- get_clauses(groups, length(arg_values)),
+         {:ok, command, values} <- find_clause(clauses, arg_values) do
+      {:ok, apply(command.handler, [message | values])}
+    end
```
Blond11516, line 51: This is so much cleaner! 😍

lib/mobius/core/command.ex
```diff
@@ -34,49 +40,44 @@ defmodule Mobius.Core.Command do
...
-      {_, {:too_few_args, _, _} = value} -> {:cont, value}
-      {:not_a_command, _} -> {:cont, :not_a_command}
-      {_, :not_a_command} -> {:cont, :not_a_command}
+  @spec preprocess_commands([t()]) :: processed()
+  def preprocess_commands(commands) do
+    commands
+    |> Enum.reverse()
+    |> Enum.group_by(fn %__MODULE__{name: name} -> name end)
+    |> Enum.map(fn {name, commands} -> {name, Enum.group_by(commands, &arg_count/1)} end)
+    |> Map.new()
```
Blond11516:
```suggestion
    |> Map.new(fn {name, commands} -> {name, Enum.group_by(commands, &arg_count/1)} end)
```
The line feels a little dense with the group_by call inside the transform but I can't think of a cleaner way to do this 🤔

Blond11516 (dismissed):

lib/mobius/cog.ex
ZeLarpMaster: From my understanding, module attributes will recalculate the value everywhere they're used, but putting it in a function means it will only be calculated once. I'm honestly still not sure about all of that, but that's what I've read while searching around. I'll go search some more to confirm or deny

lib/mobius/cog.ex
ZeLarpMaster: Sure, I thought it was better since it's inside a string interpolation, but that works too

lib/mobius/core/command.ex
ZeLarpMaster: I always forget about `new` taking a function 🤦

And yeah I agree it's a bit dense, but I tried extracting parts of that line to a function and it just felt like I was making spaghetti so I moved it back instead

lib/mobius/cog.ex
ZeLarpMaster: Looks like that was wrong:
image

Blond11516: Yeah I was pretty sure module attributes are compiled once and then inlined

Blond11516: (may not be the correct terminology but you know what I mean)

ZeLarpMaster: That's what I thought too, but when writing this PR I stumbled on something which seemed to say otherwise. I'm glad to be wrong on this, it would be annoying otherwise

Blond11516 (approved):

Issue #30, opened 2021-01-11
Caching of Discord models


ZeLarpMaster:
The following data models should be cached automatically

- [x] Guild
- [x] Member
- [x] User
- [ ] Channel
- [ ] Roles
- [ ] Emoji
- [ ] Message
- [ ] Bot Application Info

These caches are based on the list of gateway events: https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events

Issue #32, opened 2021-01-11
Write usage guides


ZeLarpMaster:
I think these guides would be relevant:

- [ ] Installation
- [ ] Discord Events
- [ ] Writing Cogs
- [ ] Interacting with Discord

If there's anything else, feel free to suggest them in the comments of this issue

PR #86, opened 2021-04-13
feat(cache): Added the guild cache

ZeLarpMaster:
I tried making the cache parse the data, but I sumbled upon a few issues. This might need a bit of discussion before doing any other cache

This PR is a part of #30 and adds the Guild cache

### DIFF
--- a/lib/mobius/actions/guild.ex
+++ b/lib/mobius/actions/guild.ex
@@ -10,16 +10,7 @@ defmodule Mobius.Actions.Guild do
   @doc """
   Gets a guild from the cache or returns `nil` if the guild was not found in the cache
 
-  ## Usage Caution
-  When using this function, please keep in mind the caches aren't perfect.
-  As stated in the
-  [Discord documentation](https://discord.com/developers/docs/reference#consistency),
-  events may never be sent to a client (this library).
-  This means the cache may not contain everything you expect it to
-  or may be out of date if an update event isn't sent (or received).
-
-  If being up to date is very important to you,
-  you may request up to date information with `get_guild/1`
+  #{Mobius.Actions.Helpers.cache_caution("get_guild/1")}
   """
   @spec get_cached_guild(Snowflake.t()) :: Guild.t() | nil
   def get_cached_guild(id), do: ModelCache.get(id, ModelCache.Guild)
--- a/lib/mobius/actions/guild.ex
+++ b/lib/mobius/actions/guild.ex
@@ -15,7 +15,7 @@ defmodule Mobius.Actions.Guild do
   As stated in the
   [Discord documentation](https://discord.com/developers/docs/reference#consistency),
   events may never be sent to a client (this library).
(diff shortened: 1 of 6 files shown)


lib/mobius/actions/guild.ex
```diff
@@ -0,0 +1,26 @@
...
+  @doc """
+  Gets a guild from the cache or returns `nil` if the guild was not found in the cache
+
+  ## Usage Caution
+  When using this function, please keep in mind the caches aren't perfect.
+  As stated in the
+  [Discord documentation](https://discord.com/developers/docs/reference#consistency),
+  events may never be sent to a client (this library).
+  This means the cache may not contain things it should contain
+  or may be out of date if an update event isn't sent (or received).
```
Blond11516:
```suggestion
  This means the cache may not contain everything you expect it to
  or may be out of date if an update event isn't sent (or received).
```

lib/mobius/actions/guild.ex
```diff
@@ -0,0 +1,26 @@
...
+  """
+
+  alias Mobius.Models.Guild
+  alias Mobius.Models.Snowflake
+  alias Mobius.Services.ModelCache
+
+  @doc """
+  Gets a guild from the cache or returns `nil` if the guild was not found in the cache
+
+  ## Usage Caution
```
Blond11516: While this is a valid warning, we're probably going to be better off moving it to a guide or something eventually to avoid repeating it for every function. If we really want to have it on all functions we should make a helper for it.

test/mobius/services/model_cache_test.exs
```diff
@@ -79,6 +79,42 @@ defmodule Mobius.Services.ModelCacheTest do
...
+    end
+
+    test "caches on guild update if already cached" do
+      original = guild()
+      cached = guild(id: original["id"])
+
+      ModelCache.cache_event(:guild_create, original)
+      ModelCache.cache_event(:guild_update, cached)
+
+      # Make sure we aren't incredibly unlucky otherwise the other assert makes no sense
```
Blond11516: What would you think of manually making them different instead of having this weird assert that requires a comment?

Blond11516 (dismissed):

lib/mobius/actions/guild.ex
ZeLarpMaster: I'd prefer if it was on every function *and* in a guide. I think there's value in having this repeated multiple times to make as many people as possible see it. This reduces the amount of people who would later come and complain about not seeing a model (user, guild, etc.) in the cache or smth. This is a very important warning and should be treated as such

I'll do a helper, the guide can come in later with #32

test/mobius/services/model_cache_test.exs
ZeLarpMaster: That's a good idea, it also removes the unlucky chance of randomly breaking this test

Blond11516 (approved):

PR #87, opened 2021-04-15
feat: Using docs in cogs and commands

ZeLarpMaster:
This PR makes `Mobius.Cog` use the `@doc` before commands and the `@moduledoc` at the top of the module for the description of the cog and the description of individual command clauses.

This feature is needed by the help command

### DIFF
--- a/lib/mobius/cog.ex
+++ b/lib/mobius/cog.ex
@@ -53,17 +53,8 @@ defmodule Mobius.Cog do
   `handle_info/2` match a different pattern.
   """
 
+  alias Mobius.Core.Cog
   alias Mobius.Core.Command
-
-  @enforce_keys [:name, :module]
-  defstruct [:name, :module, description: "", commands: []]
-
-  @type t :: %__MODULE__{
-          module: module(),
-          name: String.t(),
-          description: String.t(),
-          commands: Command.processed()
-        }
 
   @doc false
   defmacro __using__(_call) do
@@ -95,7 +86,7 @@ defmodule Mobius.Cog do
 
       @doc false
       def __cog__ do
-        %Mobius.Cog{
+        %Cog{
           module: __MODULE__,
           name: __MODULE__ |> Module.split() |> List.last(),
           description: @moduledoc,
--- a/lib/mobius/cog.ex
+++ b/lib/mobius/cog.ex
@@ -55,6 +55,16 @@ defmodule Mobius.Cog do
 
   alias Mobius.Core.Command
 
+  @enforce_keys [:name, :module]
+  defstruct name: nil, module: nil, description: "", commands: []
+
+  @type t :: %__MODULE__{
+          module: module(),
+          name: String.t(),
+          description: String.t(),
+          commands: Command.processed()
+        }
+
   @doc false
   defmacro __using__(_call) do
     quote location: :keep do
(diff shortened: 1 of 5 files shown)


lib/mobius/cog.ex
```diff
@@ -55,6 +55,16 @@ defmodule Mobius.Cog do
 
   alias Mobius.Core.Command
 
+  @enforce_keys [:name, :module]
+  defstruct name: nil, module: nil, description: "", commands: []
```
Blond11516: Why give default values to enforced keys? If it's just to avoid having to wrap the list in brackets I don't think it's worth it.

lib/mobius/cog.ex
```diff
@@ -233,30 +254,28 @@ defmodule Mobius.Cog do
...
-            arg_vars: Macro.escape(arg_vars),
+            args: Macro.escape(args),
             context: Macro.escape(context),
-            command: Macro.escape(new_command),
-            line: __CALLER__.line,
-            file: __CALLER__.file
+            module: __CALLER__.module
           ] do
-      existing_commands = Module.get_attribute(__MODULE__, :commands)
+      handler_name = Command.command_handler_name(command_name)
```
Blond11516, line 255: Why move this inside `quote`?

lib/mobius/core/command.ex
```diff
@@ -5,10 +5,11 @@ defmodule Mobius.Core.Command do
   alias Mobius.Models.Message
 
   @enforce_keys [:name, :args, :handler]
-  defstruct [:name, :args, :handler]
+  defstruct name: nil, args: nil, handler: nil, description: ""
```
Blond11516: Again, why use default values for enforced keys?

test/mobius/cog_test.exs
```diff
@@ -84,4 +84,50 @@ defmodule Mobius.CogTest do
       assert_receive {:everything, ^message, "123"}
     end
   end
+
+  describe "undocumented cog" do
+    defmodule UndocumentedCog do
+      use Mobius.Cog
+    end
+
+    test "should have nil as a description" do
```
Blond11516:
```suggestion
    test "should not have a description" do
```

Blond11516 (dismissed):

lib/mobius/cog.ex
ZeLarpMaster: I didn't know `[:name, :module, description: "", commands: []]` was valid syntax

lib/mobius/cog.ex
ZeLarpMaster: Feels simpler to reason about the macro if everything is in the quote instead of having one line outside of it and everything else inside. I can move it out if you disagree

lib/mobius/core/command.ex
ZeLarpMaster: Same answer :)

test/mobius/cog_test.exs
ZeLarpMaster: I have to disagree, `not have` sounds to me like the key doesn't exist which is different from `have nil` where the key exists and has `nil` as its value. The main difference in Elixir is if the key can *not exist* you can't do `thing.key` because that would raise a `KeyError` when it doesn't exist.

lib/mobius/cog.ex
Blond11516: I don't really care honestly. I like the idea of keeping as little stuff as possible in the quote, but I agree grouping relevant things together is nice. We can leave it as is.

test/mobius/cog_test.exs
Blond11516: I would argue the difference between `nil` and not having the key at all is an implementation detail, which therefore should not be reflected in the test name.

Blond11516 (approved):

PR #90, opened 2021-04-18
feat(actions): Implement edit channel action

Blond11516:
Implements the `edit_channel` action. Also adds some utility functions for validating action parameters.

This PR is part of #29.

### DIFF
--- a/lib/mobius/actions/channel.ex
+++ b/lib/mobius/actions/channel.ex
@@ -3,11 +3,15 @@ defmodule Mobius.Actions.Channel do
   Actions related to Discord channels such as fetching, modifying and deleting channels
   """
 
+  import Mobius.Actions.Utils
+
   alias Mobius.Models.Channel
   alias Mobius.Models.Snowflake
   alias Mobius.Rest
   alias Mobius.Rest.Client
   alias Mobius.Services.Bot
+
+  @type edit_channel_body :: Rest.Channel.edit_channel_body()
 
   @doc """
   Fetch a channel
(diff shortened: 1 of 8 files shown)


lib/mobius/actions/channel.ex
```diff
@@ -26,4 +30,31 @@ defmodule Mobius.Actions.Channel do
   def get_channel(channel_id) do
     Rest.Channel.get_channel(Bot.get_client!(), channel_id)
   end
+
+  @spec edit_channel(pos_integer, edit_channel_body()) :: Client.result(Channel.t())
```
ZeLarpMaster: channel_id's type should be `Snowflake.t()`

lib/mobius/actions/utils.ex
```diff
@@ -0,0 +1,61 @@
...
+
+  @spec string_length_validator(Map.key(), non_neg_integer(), non_neg_integer()) ::
+          validator()
+  def string_length_validator(key, min, max) do
+    fn
+      %{^key => value} when not is_binary(value) ->
+        {:error, "Expected #{key} to be a string, got #{inspect(value)}"}
+
+      %{^key => value} ->
+        string_length = String.length(value)
```
ZeLarpMaster: I wonder if Discord counts graphemes or codepoints. It would do a difference if, for example, there's emojis or multi-codepoint characters in channel names

test/mobius/actions/utils_test.exs
```diff
@@ -0,0 +1,91 @@
...
+  describe "string_length_validator/3" do
+    setup do
+      min = 2
+      max = 4
+      validator = Utils.string_length_validator(:foo, min, max)
+
+      [min: min, max: max, validator: validator]
+    end
+
+    test "validator returns an error when provided a non-string input", ctx do
```
ZeLarpMaster: saying `validator` is a bit redundant with the describe; applies to all tests in this file

test/support/fixtures.ex
```diff
@@ -119,6 +119,15 @@ defmodule Mobius.Fixtures do
 
   def random_discriminator, do: random_n_chars(4, "0123456789")
 
+  def random_text(len) do
+    alphabet =
+      ?a..?z
+      |> Enum.to_list()
+      |> List.to_string()
```
ZeLarpMaster: Wouldn't it be simpler to just hardcode `"abcdefghijklmnopqrstuvwxyz"`?

lib/mobius/actions/utils.ex
```diff
@@ -0,0 +1,61 @@
+defmodule Mobius.Actions.Utils do
```
ZeLarpMaster: While this is indeed useful, that's kind of the point of the `Mobius.Validation` modules. So I think you could just move this in there

ZeLarpMaster (changes requested):

lib/mobius/actions/utils.ex
Blond11516: Just did a few tests. The API seems to count graphemes (generated 1024 characters, replaced 1 with an emoji and it worked, the client even displayed it properly), but if you try to do it from the client, it converts them to their textual representation (ie. `:thinking:`), which takes up more characters.

Given those results, using `String.length` appears to be the correct way here.

test/support/fixtures.ex
Blond11516: Yeah probably, I guess I got carried away :sweat_smile:

lib/mobius/actions/utils.ex
ZeLarpMaster: Alright, thank you for trying that out :)

ZeLarpMaster (approved):
Good stuff! This will help with future actions too which is very nice :)

PR #94, opened 2021-05-06
feat: Using command returns to reply

### LINKED ISSUE #91: feat: Use commands' return value
Commands' return value could be one of:
- `:ok`
- `{:react, emoji}`
- `{:reply, message}`
- `{:private_reply, message}`

Their names are pretty self-explanatory. For example, `{:reply, message}` just maps to `send_message(message, context.channel_id)`

This would be done in the `listen :message_create` generated in `Mobius.Cog.__before_compile__` (or ideally done in a function called from there).

For now we could support only `:ok` and add support for the others in a later PR (or all at once if you feel motivated) since this introduces a breaking change to the command API which should be addressed as soon as possible.

ZeLarpMaster:
Closes #91

This PR adds some testing utilities for cogs. I'm fairly happy with the API for it, though the implementation is a bit shitty because of how Tesla.Mock works. I think that would be worth fixing after the v1.

This PR also includes a fix on the REST client user agent missing a `)`

### DIFF
--- a/lib/mobius/cog.ex
+++ b/lib/mobius/cog.ex
@@ -53,6 +53,8 @@ defmodule Mobius.Cog do
   `handle_info/2` match a different pattern.
   """
 
+  import Mobius.Actions.Message
+
   alias Mobius.Core.Cog
   alias Mobius.Core.Command
 
@@ -96,8 +98,8 @@ defmodule Mobius.Cog do
 
       listen :message_create, message do
         case Command.execute_command(@computed_commands, Bot.get_global_prefix!(), message) do
-          {:ok, _} ->
-            :ok
+          {:ok, value} ->
+            Mobius.Cog.handle_return(value, message)
 
           {:too_few_args, arities, received} ->
             Logger.info(
@@ -289,6 +291,10 @@ defmodule Mobius.Cog do
       command(unquote(command_name), _, [], do: unquote(block))
     end
   end
+
+  @doc false
+  def handle_return(:ok, _context), do: :ok
+  def handle_return({:reply, body}, context), do: send_message(body, context.channel_id)
 
   @doc false
   def __event_handler_name__(event_name, event_handlers) do
--- a/lib/mobius/cog.ex
+++ b/lib/mobius/cog.ex
@@ -58,6 +58,8 @@ defmodule Mobius.Cog do
   alias Mobius.Core.Cog
   alias Mobius.Core.Command
 
+  require Logger
+
   @doc false
   defmacro __using__(_call) do
     quote location: :keep do
@@ -295,6 +297,7 @@ defmodule Mobius.Cog do
   @doc false
   def handle_return(:ok, _context), do: :ok
   def handle_return({:reply, body}, context), do: send_message(body, context.channel_id)
+  def handle_return(value, _context), do: Logger.warn("Invalid return: #{inspect(value)}")
 
   @doc false
   def __event_handler_name__(event_name, event_handlers) do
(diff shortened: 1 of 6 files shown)


lib/mobius/cog.ex
```diff
@@ -290,6 +294,11 @@ defmodule Mobius.Cog do
     end
   end
 
+  @doc false
+  def handle_return(:ok, _context), do: :ok
+  def handle_return({:reply, body}, context), do: send_message(body, context.channel_id)
+  def handle_return(value, _context), do: Logger.warn("Invalid return: #{inspect(value)}")
```
ZeLarpMaster: Had to add this so I could test it. I didn't find a way to detect the cog crashing (I tried `Process.monitor` on the cog, but it crashed without notifying the test :/ ). If you have an idea, I'm open to changing this, I don't really like having to wait for logs.

test/support/cog_stub.ex
```diff
@@ -37,6 +37,14 @@ defmodule Mobius.Stubs.Cog do
     send_to_test({:unexpected_everything, value})
   end
 
+  command "unsupported" do
+    :unsupported_return
+  end
+
+  command "reply" do
+    {:reply, %{content: "The answer"}}
```
ZeLarpMaster, line 45: Would it be worth changing the other tests which use the cog_stub so they do this too? We wouldn't need the whole `:cog_test_process` thing I think

test/support/cog_test_utils.ex
```diff
@@ -0,0 +1,33 @@
...
+    mock_global(fn env ->
+      send(test_pid, env)
+      {200, [], %{}}
+    end)
+  end
+
+  @spec assert_message_sent(Message.message_body()) :: :ok
+  def assert_message_sent(message) do
+    json_message = Jason.encode!(message)
+    %Tesla.Env{url: url} = assert_receive %Tesla.Env{body: ^json_message}
```
ZeLarpMaster: Should assert it's a `:post`

lib/mobius/cog.ex
Blond11516: I agree that doesn't sound ideal, but I have no idea how people usually do this? :thinking: Good enough for now I guess :shrug:

test/support/cog_stub.ex
Blond11516: Probably in another PR, but yeah I think that would be nice!

test/mobius/cog_test.exs
```diff
@@ -85,6 +86,24 @@ defmodule Mobius.CogTest do
...
+
+      assert_message_sent(%{content: "The answer"})
+    end
+
+    test "should raise an error if the command returns something unsupported" do
+      assert capture_log(fn ->
+               send_command_payload("unsupported")
+               # Give time for the log to appear
+               Process.sleep(10)
+             end) =~ "Invalid return: :unsupported_return"
```
Blond11516: What if you actually raised? Couldn't you do
```elixir
try do
  send_command_payload("unsupported")
rescue
  e -> assert ...
end
```

Blond11516 (dismissed):

test/mobius/cog_test.exs
ZeLarpMaster: The exception isn't raised in the same process. `send_command_payload` is an indirect `send/2` to the Cog process so we wouldn't be able to get it here. I tried monitoring the process, but that also didn't seem to work. I'll try to investigate that though, I think it'd be useful if we could do that.

ZeLarpMaster: Figured it out and it seems to work! It looks better though it doesn't assert *why* the cog dies (would be hard and invasive on the implementation details), but it's something.

Writing this made me realize a mixed approach would be even better. We can capture the log and look for the error logged by the unhandled FunctionClauseError and assert on that too 😄

Blond11516 (approved):

PR #97, opened 2021-05-15
feat: Supporting stage channels

### LINKED ISSUE #85: feat: Support the new Stage channels
A new channel type has appeared: [Stage Channels](https://support.discord.com/hc/en-us/articles/1500005513722-Stage-Channels-FAQ)

We should do necessary changes to `Mobius.Models.Channel` to support them

ZeLarpMaster:
Made by looking at the git history of the api docs and looking for commits which mention "stage".

I did not add the CRUD operations and will instead add them when the need comes.

I also disabled the `SinglePipe` rule in credo. I originally thought it's a good idea, but it's rubbing me the wrong way, especially when I expect new pipes to be added in the future like in this case (you'll see where when reviewing).

Closes #85


Blond11516 (approved):

Blond11516:
I may be blind but I didn't find any instance of a single pipe :thinking:

lib/mobius/rest/channel.ex
```diff
@@ -28,8 +30,19 @@ defmodule Mobius.Rest.Channel do
 
   @spec edit_channel(Client.t(), Snowflake.t(), edit_channel_body()) :: Client.result(Channel.t())
   def edit_channel(client, channel_id, params) do
+    params =
+      params
+      # TODO: |> update_param(:type, &convert_type/1)
+      |> update_param(:video_quality_mode, &convert_video_quality_mode/1)
```
ZeLarpMaster: It was here, but got removed because `Mobius.Rest.Channel` was deleted when I merged rewrite into this branch

ZeLarpMaster: There's places where we could use these to improve clarity, I just didn't want to run around the whole project and change them in this PR

PR #103, opened 2021-08-07
refactor: Use endpoint generator for list messages action

Blond11516:
Adds the `Endpoint` module and the action generator. In the future, this should help streamline the implementation of new endpoints, making development and maintenance much easier.

### DIFF
--- a/lib/mobius/actions.ex
+++ b/lib/mobius/actions.ex
@@ -3,7 +3,7 @@ defmodule Mobius.Actions do
   Generation of Mobius actions
 
   This module contains the macros and functions necessary to generate Mobius
-  actions form endpoint definitions.
+  actions from endpoint definitions.
 
   ## Warning about exported functions
 
--- a/lib/mobius/actions.ex
+++ b/lib/mobius/actions.ex
@@ -1,0 +1,98 @@
+defmodule Mobius.Actions do
+  @moduledoc """
+  Generation of Mobius actions
+
+  This module contains the macros and functions necessary to generate Mobius
+  actions form endpoint definitions.
+
+  ## Warning about exported functions
+
+  The functions exported by this module are not meant to be used directly. They
+  are used by the `setup_actions/1` macro, which should be considered to be the
+  sole entrypoint for this module.
+  """
+
+  alias Mobius.Endpoint
+  alias Mobius.Services.Bot
+  alias Mobius.Validations.ActionValidations
+
+  @doc """
+  Generates actions functions from a list of `Mobius.Endpoint`.
+  """
+  @spec setup_actions([Endpoint.t()]) :: Macro.output()
+  defmacro setup_actions(endpoints) do
+    quote bind_quoted: [endpoints: endpoints], location: :keep do
+      import Mobius.Validations.ActionValidations
+
+      alias Mobius.Actions
+
+      resource = __MODULE__ |> Module.split() |> List.last()
+
+      Enum.each(endpoints, fn endpoint ->
+        name = endpoint.name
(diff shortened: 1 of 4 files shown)


lib/mobius/actions.ex
```diff
@@ -0,0 +1,98 @@
+defmodule Mobius.Actions do
+  @moduledoc """
+  Generation of Mobius actions
+
+  This module contains the macros and functions necessary to generate Mobius
+  actions form endpoint definitions.
```
ZeLarpMaster: `form` --> `from`

lib/mobius/actions.ex
```diff
@@ -0,0 +1,98 @@
...
+            unquote(resource)
+          )
+        end
+      end)
+    end
+  end
+
+  @spec execute(Endpoint.t(), Keyword.t(any()), atom() | binary()) :: any()
+  def execute(%Endpoint{} = endpoint, params, resource) do
+    rest_module = Module.concat([:Mobius, :Rest, resource])
```
ZeLarpMaster, line 49: `resource` is only used to do this, why not pass this module directly instead of doing `Module.split` and then `Module.concat`?

lib/mobius/actions/message.ex
```diff
@@ -5,11 +5,17 @@ defmodule Mobius.Actions.Message do
...
+  alias Mobius.Actions
   alias Mobius.Models.Message
+  alias Mobius.Models.Snowflake
   alias Mobius.Rest
   alias Mobius.Rest.Client
   alias Mobius.Services.Bot
 
+  require Actions
+
+  Actions.setup_actions(Mobius.Endpoint.Message.endpoints())
```
ZeLarpMaster, line 17: Not sure I like the endpoint being defined in another module. Especially since `send_message` isn't defined through this, it means you need to know and remember for each endpoint where it's defined instead of having it all in the same module.
I'd rather have the `endpoints` function in this same module instead

lib/mobius/endpoint.ex
```diff
@@ -0,0 +1,25 @@
...
+  @moduledoc """
+  An endpoint that can be used to interact with Discord
+
+  The attributes of an endpoint are the following:
+  - name: The name of the endpoint. This will correspond to the name of the
+  exposed by Mobius to query this endpoint.
+  - url: The url at which Discord exposes the endpoint.
+  - method: The HTTP method to use when sending requests to Discord.
+  - params: The list of query parameters the endpoint accepts.
+  - opts: The options that the endpoint accpets. These generaly correspond to
```
ZeLarpMaster: `accpets` --> `accepts`
`generaly` --> `generally`

lib/mobius/actions.ex
Blond11516: Because it's not the same module. We're going from `Mobius.Actions.Message` to `Mobius.Rest.Message`

lib/mobius/actions/message.ex
Blond11516: Yeah, my only concern is that I was hoping to write a generator for the `Rest` modules as well, which I feel is gonna be kind of awkward if the rest layer has to pull endpoints from the actions layer.

lib/mobius/actions.ex
ZeLarpMaster: Ah, true, I forgot `resource` is `__MODULE__` as I was reading through

lib/mobius/actions/message.ex
ZeLarpMaster: Ah I see, then it's unfortunate I guess 🤷

ZeLarpMaster (approved):

PR #104, opened 2021-08-11
refactor: Create REST engine to abstract Tesla rest handlers

Blond11516:
Creates a `Mobius.Rest.execute/2` function that can execute a variety of different HTTP calls. With this, the whole action pipeline should be fully automated for most actions.


lib/mobius/endpoint.ex
```diff
@@ -10,16 +10,25 @@ defmodule Mobius.Endpoint do
...
 
   @type t :: %__MODULE__{
           name: atom(),
           url: String.t(),
           method: :get | :post,
           params: [atom()],
-          opts: map()
+          opts: map() | nil,
+          model: atom() | nil,
+          list_response?: boolean() | nil
```
ZeLarpMaster, line 28: Pro tip: to avoid the `- opts: map()` and `+ opts: map()` in the git diff, you can add new params to your struct in the fields before the last one. So if you put `model` and `list_response?` before `opts`, you only get the true diff

lib/mobius/rest.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Mobius.Rest do
+  @moduledoc false
+
+  alias Mobius.Endpoint
+  alias Mobius.Rest.Client
+
+  @spec execute(Endpoint.t(), Tesla.Client.t(), keyword()) :: Client.result(any())
+  def execute(%Endpoint{} = endpoint, client, args) do
```
ZeLarpMaster, line 8: Not sure whether this function deserves some tests or we just let it be tested through tests for other modules 🤔

ZeLarpMaster (approved):

lib/mobius/endpoint.ex
Blond11516: pro tip indeed! This time I did add ` | nil` to the opts type but I'll keep this in mind in the future!

lib/mobius/rest.ex
Blond11516: Good question 🤔 Since we're basically going to get rid of almost all other rest modules I guess this would have to be tested through actions, but we're already pretty much doing that so I think it should be fine?

PR #105, opened 2021-08-14
refactor: Make action validators take values directly

Blond11516:
Refactors the action validators to receive the value to validate directly instead of extracting it from the map of options.

This will make it far easier to implement a `list_validator`, which we will need for future endpoints.


ZeLarpMaster (approved):
Looks good to me 👍

PR #106, opened 2021-08-15
refactor: Use action generator for channel endpoints

Blond11516:
* Replace manually implemented channel actions with automatically generated ones.
* Add support for custom validators in the `ActionValidators` module.


lib/mobius/validations/action_validations.ex
```diff
@@ -87,6 +93,7 @@ defmodule Mobius.Validations.ActionValidations do
   def get_validator({:integer, opts}), do: get_integer_range_validator(opts)
   def get_validator(:string), do: string_validator()
   def get_validator({:string, opts}), do: get_string_length_validator(opts)
+  def get_validator({module, function}), do: fn val -> apply(module, function, [val]) end
```
Blond11516, line 96:
```suggestion
  def get_validator(validator), do: validator
```
I just checked, we _could_ have this take a function directly, but the custom validators would have to be defined outside of the the actions module in order to be able to capture them in the endpoint definition. Not sure which one is better :thinking:

ZeLarpMaster: Ah because it's a macro the module itself isn't defined yet when the macro is executed, that's unfortunate

I prefer having the option to put the custom validator in the module itself, makes it a little bit less spaghetti I guess

ZeLarpMaster (approved):
I don't know how we'll validate permissions with this, but I guess we'll figure that out when we get there

Blond11516:
Yeah I haven't given this that much thought yet but I figured a list of required permissions in the endpoint should probably be a good starting point 🤷

PR #107, opened 2021-08-17
feat: Validate types of all action parameters

Blond11516:
Uses `ActionValidations` to validate the types of all action parameters.

### DIFF
--- a/lib/mobius/actions/channel.ex
+++ b/lib/mobius/actions/channel.ex
@@ -18,14 +18,14 @@ defmodule Mobius.Actions.Channel do
       name: :get_channel,
       url: "/channels/:channel_id",
       method: :get,
-      params: [:channel_id],
+      params: [{:channel_id, :snowflake}],
       model: Mobius.Models.Channel
     },
     %Endpoint{
       name: :edit_channel,
       url: "/channels/:channel_id",
       method: :patch,
-      params: [:channel_id],
+      params: [{:channel_id, :snowflake}],
       opts: %{
         name: {:string, [min: 2, max: 100]},
         type: {__MODULE__, :validate_channel_type},
@@ -40,7 +40,7 @@ defmodule Mobius.Actions.Channel do
       name: :delete_channel,
       url: "/channels/:channel_id",
       method: :delete,
-      params: [:channel_id],
+      params: [{:channel_id, :snowflake}],
       model: Mobius.Models.Channel
     }
   ])
--- a/lib/mobius/validations/action_validations.ex
+++ b/lib/mobius/validations/action_validations.ex
@@ -73,7 +73,7 @@ defmodule Mobius.Validations.ActionValidations do
         val = params[param_name]
 
         if val != nil do
-          case validator.(params[param_name]) do
+          case validator.(val) do
             :ok -> errors
             {:error, error} -> ["Expected #{param_name} to #{error}" | errors]
           end
--- a/lib/mobius/validations/action_validations.ex
+++ b/lib/mobius/validations/action_validations.ex
@@ -66,11 +66,13 @@ defmodule Mobius.Validations.ActionValidations do
     end
   end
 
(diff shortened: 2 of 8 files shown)


lib/mobius/actions/channel.ex
```diff
@@ -18,14 +18,14 @@ defmodule Mobius.Actions.Channel do
...
-      params: [:channel_id],
+      params: [{:channel_id, :snowflake}],
       model: Mobius.Models.Channel
     },
     %Endpoint{
       name: :edit_channel,
       url: "/channels/:channel_id",
       method: :patch,
-      params: [:channel_id],
+      params: [{:channel_id, :snowflake}],
```
ZeLarpMaster, line 28: Makes me sad that this is useful just because [webhook-related calls use tokens in the path](https://discord.com/developers/docs/interactions/receiving-and-responding#create-interaction-response) which means *sometimes* it won't be a snowflake. The only two instances where the path params aren't snowflakes is for application commands (previously known as slash commands) and webhooks (which we won't support anyway)

ZeLarpMaster: I would've preferred to have `:snowflake` be a default and needing to specify only if it isn't, but it would make this whole thing a lot harder to implement and maintain :(

lib/mobius/validations/action_validations.ex
```diff
@@ -66,11 +66,13 @@ defmodule Mobius.Validations.ActionValidations do
...
-  @spec validate_params(map(), [{atom(), validator()}]) :: :ok | {:error, [String.t()]}
-  def validate_params(params, validators) do
+  @spec validate_args(Access.t(), [{atom(), validator()}]) :: :ok | {:error, [String.t()]}
+  def validate_args(params, validators) do
     errors =
       Enum.reduce(validators, [], fn {param_name, validator}, errors ->
-        if Map.has_key?(params, param_name) do
+        val = params[param_name]
+
+        if val != nil do
```
ZeLarpMaster, line 75: Are you using `val` other than in this `if`? Why this change?

Blond11516: By using access instead of `Map.get` we can accept either a map or keyword list and it just so happens that with the changes to `Actions` it became more convenient to pass a keyword list.

Also I was intending to pass `val` to the validator at line 76 (instead of doing a second access) but I forgot, I'll fix it 😛

lib/mobius/actions/channel.ex
Blond11516: It's true that having `:snowflake` be the default would be a lot simpler 🤔 An easy way to support that would be to update the params to add the type before doing anything else in `Actions.execute`:
```elixir
Enum.map(
  endpoint.params,
  fn
    {name, type} -> {name, type}
    name -> {name, :snowflake}
  end
)
```
The main issue I see with this solution is that the type for endpoints would have to allow both forms, but _everything_ would assume the tuple form except for the actual endpoint definitions, which really isn't the clearest thing ever 😕

Blond11516: Another thing we could do is keep the endpoint struct as-is (with the typed parameters) but add a `new` function which would handle the default value:
```elixir
# Mobius.Endpoint

def new(endpoint) do
  params = Enum.map(
    endpoint[:params],
    fn
      {name, type} -> {name, type}
      name -> {name, :snowflake}
    end
  )

  struct(__MODULE__, %{endpoint | params: params})
end
```
It does mean we would have to almost exactly duplicate the `Endpoint.t()` type to typespec this function but I still think it would be much more maintainable this way. What do you think?

ZeLarpMaster: I think I prefer having to write (and read) `{, :snowflake}` once per path param than having to deal with something which injects default values. I wanna keep things as simple as they can realistically be for the sake of this project staying maintainable. I know this isn't much, but little things add up quick and I really don't mind the `{, :snowflake}`. It's unfortunate that Discord has anything other than snowflakes in path params 😕

ZeLarpMaster (approved):

PR #108, opened 2021-08-22
feat: Add support for documentation generation in actions

Blond11516:
Adds support for generating function documentation in the action generator. Documentations include a (mandatory) link to the official discord doc as well as an optional custom section.

### DIFF
--- a/lib/mobius/actions.ex
+++ b/lib/mobius/actions.ex
@@ -30,7 +30,9 @@ defmodule Mobius.Actions do
         name = endpoint.name
         param_names = Actions.get_arguments(endpoint)
         params_keyword = Actions.get_arguments_keyword(endpoint)
+        docs = Actions.get_docs(endpoint)
 
+        @doc docs
         def unquote(name)(unquote_splicing(param_names)) do
           Actions.execute(
             unquote(Macro.escape(endpoint)),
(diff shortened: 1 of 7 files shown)


lib/mobius/actions.ex
```diff
@@ -30,7 +30,11 @@ defmodule Mobius.Actions do
         name = endpoint.name
         param_names = Actions.get_arguments(endpoint)
         params_keyword = Actions.get_arguments_keyword(endpoint)
+        docs = Actions.get_docs(endpoint)
 
+        @doc """
+        #{docs}
+        """
```
ZeLarpMaster: `@doc docs` doesn't do the same thing?

test/mobius/actions/message_test.exs
```diff
@@ -28,7 +28,7 @@ defmodule Mobius.Actions.MessageTest do
       channel_id = random_snowflake()
       raw = message(channel_id: channel_id)
       url = Client.base_url() <> "/channels/#{channel_id}/messages"
-      mock(fn %{method: :post, url: ^url} -> json(raw) end)
+      mock(fn %{method: :post, url: ^url, body: _} -> json(raw) end)
```
ZeLarpMaster: why do we need `body: _`?

lib/mobius/actions.ex
Blond11516: Yup, it does :+1:

test/mobius/actions/message_test.exs
Blond11516: I honestly don't remember :thinking: Feels like it had to with making sure we were passing the body to tesla, but tesla doesn't even provide a way to do a PATCH, POST or PUT without a body...

I've removed it.

ZeLarpMaster (approved):

PR #111, opened 2021-09-06
feat: Add get_message action

Blond11516:
Implements the `get_message` action. This is part of https://github.com/ZeLarpMaster/Mobius/issues/29.

### DIFF
--- a/lib/mobius/actions/message.ex
+++ b/lib/mobius/actions/message.ex
@@ -47,6 +47,23 @@ defmodule Mobius.Actions.Message do
           {:ok, [%Mobius.Models.Message{} = message]}
       """,
       model: Mobius.Models.Message
+    },
+    %Endpoint{
+      name: :get_message,
+      url: "/channels/:channel_id/messages/:message_id",
+      method: :get,
+      params: [{:channel_id, :snowflake}, {:message_id, :snowflake}],
+      discord_doc_url:
+        "https://discord.com/developers/docs/resources/channel#get-channel-message",
+      doc: """
+      Fetches a single message in a channel
+
+      ## Example
+
+          iex> get_message("123456789", "987654321")
+          {:ok, %Mobius.Molels.Message{} = message}
+      """,
+      model: Mobius.Models.Message
     }
   ])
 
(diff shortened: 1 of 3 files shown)


lib/mobius/actions/message.ex
```diff
@@ -47,6 +47,23 @@ defmodule Mobius.Actions.Message do
...
+      params: [{:channel_id, :snowflake}, {:message_id, :snowflake}],
+      discord_doc_url:
+        "https://discord.com/developers/docs/resources/channel#get-channel-message",
+      doc: """
+      Fetches a single message in a channel
+
+      ## Example
+
+          iex> get_message("123456789", "987654321")
+          {:ok, %Mobius.Molels.Message{} = message}
```
ZeLarpMaster: `Molels` 😂 that's a funny typo

test/mobius/actions/message_test.exs
```diff
@@ -101,4 +101,32 @@ defmodule Mobius.Actions.MessageTest do
...
+      message_id = random_snowflake()
+      raw = message(channel_id: channel_id, id: message_id)
+      url = Client.base_url() <> "/channels/#{channel_id}/messages/#{message_id}"
+      mock(fn %{method: :get, url: ^url} -> json(raw) end)
+      [channel_id: channel_id, message_id: message_id, raw_message: raw]
+    end
+
+    test "returns an error if channel_id is not a snowflake", ctx do
+      {:error, errors} = Message.get_message(1324, ctx.message_id)
+      assert_has_error(errors, "Expected channel_id to be a snowflake")
```
ZeLarpMaster, line 119: I'm wondering whether snowflake params should be integers or strings. When we parse snowflakes from return values, we return it as an integer and say its type is a snowflake so I feel like we should do the opposite for params (take integer and send to discord as strings) so users of the lib always need to send integers when a snowflake is expected.

If there's anything to change, I don't think it should be changed in this pr, but that might be something to keep in mind. Also completely my bad for not spotting this earlier

ZeLarpMaster (dismissed):

lib/mobius/actions/message.ex
Blond11516: Woops 😅

test/mobius/actions/message_test.exs
Blond11516: Huh, for some reason I actually thought we went out of our way to use strings instead of integers 🤔 I agree we should change it to be more consistent 👍 I'll open an issue.

ZeLarpMaster (approved):

PR #113, opened 2021-09-15
feat: Implement the delete_own_reaction action

Blond11516:
- Implements the `delete_own_reaction` action.
- Adds a simple argument preprocessor to actions. Right now it is only used to get the identifier from an emoji.
- Adds an emoji action validator

Part of #29

### DIFF
--- a/lib/mobius/actions.ex
+++ b/lib/mobius/actions.ex
@@ -13,6 +13,7 @@ defmodule Mobius.Actions do
   """
 
   alias Mobius.Endpoint
+  alias Mobius.Models.Emoji
   alias Mobius.Services.Bot
   alias Mobius.Validations.ActionValidations
 
@@ -51,7 +52,8 @@ defmodule Mobius.Actions do
 
     case ActionValidations.validate_args(params ++ path_params, validators) do
       :ok ->
-        Mobius.Rest.execute(endpoint, Bot.get_client!(), params)
+        processed_params = pre_process_params(endpoint, params)
+        Mobius.Rest.execute(endpoint, Bot.get_client!(), processed_params)
 
       {:error, errors} ->
         {:error, errors}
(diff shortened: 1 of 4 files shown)


lib/mobius/actions.ex
```diff
@@ -104,4 +106,21 @@ defmodule Mobius.Actions do
           {atom(), ActionValidations.validator()}
   defp type_tuple_to_validator_tuple({name, type}),
     do: {name, ActionValidations.get_validator(type)}
+
+  @spec pre_process_params(Endpoint.t(), keyword()) :: keyword()
+  defp pre_process_params(%Endpoint{params: params}, params_values) do
+    Enum.map(params_values, fn {name, value} ->
+      processed_value =
+        params[name]
+        |> then(fn type -> pre_process_param_value(type, value) end)
```
ZeLarpMaster: Or `processed_value = pre_process_param_value(params[name], value)`

Do we get something by using `then` here? Feels a bit like too much to me

lib/mobius/actions.ex
```diff
@@ -104,4 +106,21 @@ defmodule Mobius.Actions do
...
+
+      {name, processed_value}
+    end)
+  end
+
+  @spec pre_process_param_value(ActionValidations.validator_type(), any()) :: any()
+  defp pre_process_param_value(:emoji, %Emoji{} = emoji),
+    do: Emoji.get_identifier(emoji)
+
+  defp pre_process_param_value(_, value), do: value
```
ZeLarpMaster, line 123: Alternatively maybe we could use a [Protocol](https://elixir-lang.org/getting-started/protocols.html) instead?

lib/mobius/actions.ex
Blond11516: Good point. I think I initially had it as part of a bigger pipe, not sure :thinking:

lib/mobius/actions.ex
Blond11516: We could, although I'm not quite sure what we have to gain here.
In my mind protocols are great for dependency injection-type stuff (i.e. letting the user provide their own implementation). I don't think this is such a case, though maybe there's another advantage I'm not seeing.

ZeLarpMaster: Fair point 🤔 It would indeed be pretty bloated for what it's worth here

ZeLarpMaster (approved):

PR #117, opened 2021-09-26
refactor: Create a Model behaviour

Blond11516:
- Renames `Mobius.Models.Utils` to `Mobius.Model`
- Adds a callback for the `parse/1` function
- Replace `apply/3` call with the new callback


ZeLarpMaster (approved):
Very neat! I like it

PR #121, opened 2021-10-24
feat: Implement the edit_message action

Blond11516:
Part of #29

- Adds the edit_message action.
- Adds the notion of "constraints" to endpoints. Constraints are like validations, except that they can look at multiple parameters at a time instead of only one. Useful for implementing "at least on of" or "only one of" type validations.
- Adds support for multipart messages to the generic rest module. It currently assumes a `:file` option. Looking at the discord docs, it looks like `create_message` and `edit_message` are the only two endpoints that use such an option.
- Documents the different available validations and constraints for actions
- Adds a step to the action executor to check that the bot is ready.
- In the `hanshake_shard` test setup, add a small delay before returning to make sure the shard has had time to receive and handle the message telling it it's ready. Also asserts that the bot is ready.
- Allow the `assert_has_error` test helper to receive only a single error instead of a list.

### DIFF
--- a/lib/mobius/actions/message.ex
+++ b/lib/mobius/actions/message.ex
@@ -64,6 +64,40 @@ defmodule Mobius.Actions.Message do
           {:ok, %Mobius.Models.Message{} = message}
       """,
       model: Mobius.Models.Message
+    },
+    %Endpoint{
+      name: :edit_message,
+      url: "/channels/:channel_id/messages/:message_id",
+      method: :patch,
+      params: [{:channel_id, :snowflake}, {:message_id, :snowflake}],
+      opts: %{
+        content: {:string, [max: 2000]},
+        embed: :any,
+        flags: :integer,
+        file: :any,
+        allowed_mentions: :any
+      },
+      constraints: [{:one_of, [:content, :embed, :file]}],
+      discord_doc_url: "https://discord.com/developers/docs/resources/channel#edit-message",
+      model: Mobius.Models.Message,
+      multipart?: true,
+      doc: """
+      Edits an exsting message
+
+      Only the original author of the message may edit the "content" and
+      "embeds" fields. Flags may be edited by anyone.
+
+      Be awarae that updating flags works as a complete replacement. This means
+      that if you want to add a new flag, you also need to set all existing
+      flags along with the new one.
+
+      Refer to `send_message/2` for more information on the message body format.
+
+      ## Example
+
+          iex> send_message(%{content: "Some content"}, channel_id)
+          {:ok, %Mobius.Models.Message{} = message}
+      """
     }
   ])
 
--- a/lib/mobius/actions/message.ex
+++ b/lib/mobius/actions/message.ex
@@ -77,17 +77,17 @@ defmodule Mobius.Actions.Message do
         file: :any,
         allowed_mentions: :any
       },
-      constraints: [{:one_of, [:content, :embed, :file]}],
+      constraints: [{:at_least_one_of, [:content, :embed, :file]}],
       discord_doc_url: "https://discord.com/developers/docs/resources/channel#edit-message",
       model: Mobius.Models.Message,
       multipart?: true,
       doc: """
-      Edits an exsting message
+      Edits an existing message
 
       Only the original author of the message may edit the "content" and
       "embeds" fields. Flags may be edited by anyone.
 
-      Be awarae that updating flags works as a complete replacement. This means
+      Be aware that updating flags works as a complete replacement. This means
       that if you want to add a new flag, you also need to set all existing
       flags along with the new one.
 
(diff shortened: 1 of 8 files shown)


lib/mobius/actions/message.ex
```diff
@@ -64,6 +64,40 @@ defmodule Mobius.Actions.Message do
...
+      method: :patch,
+      params: [{:channel_id, :snowflake}, {:message_id, :snowflake}],
+      opts: %{
+        content: {:string, [max: 2000]},
+        embed: :any,
+        flags: :integer,
+        file: :any,
+        allowed_mentions: :any
+      },
+      constraints: [{:one_of, [:content, :embed, :file]}],
```
ZeLarpMaster: I think `:at_least_one_of` would be more appropriate since I'd expect an `:only_one_of` to happen later and it would cause confusion

lib/mobius/actions/message.ex
```diff
@@ -64,6 +64,40 @@ defmodule Mobius.Actions.Message do
...
+      discord_doc_url: "https://discord.com/developers/docs/resources/channel#edit-message",
+      model: Mobius.Models.Message,
+      multipart?: true,
+      doc: """
+      Edits an exsting message
```
ZeLarpMaster: Typo: `existing`

lib/mobius/actions/message.ex
```diff
@@ -64,6 +64,40 @@ defmodule Mobius.Actions.Message do
...
+
+      Only the original author of the message may edit the "content" and
+      "embeds" fields. Flags may be edited by anyone.
+
+      Be awarae that updating flags works as a complete replacement. This means
```
ZeLarpMaster: Typo: `aware`

lib/mobius/endpoint.ex
```diff
@@ -17,6 +17,10 @@ defmodule Mobius.Endpoint do
   - discord_doc_url: The URL for the official Discord documentation for this
   endpoint.
   - doc: The documentation of the function.
+  - multipart?: Whether the endpoint can contain multipart form data. If true,
+  the endpoint has to have an option with the name `:file`.
+  - constraints: Additional contstraints for `:opts` that that involve
```
ZeLarpMaster: Typo: `contstraints` --> `constraints`

lib/mobius/validations/action_validations.ex
```diff
@@ -1,17 +1,75 @@
...
+  a one-arity function in that module. The function will receive the option
+  value as its only argument.
+
+  ### :emoji
+
+  An emoji. Must be a `Mobius.Models.Emoji` struct.
+
+  ## Constraints
+
+  Contsraints are similar to validator in that they verify that some conditions
```
ZeLarpMaster: Typo: `Constraints`

lib/mobius/validations/action_validations.ex
```diff
@@ -71,6 +129,32 @@ defmodule Mobius.Validations.ActionValidations do
...
+          options
+          |> Enum.map(fn option -> params[option] end)
+          |> Enum.any?(fn val -> val != nil end)
+
+        if valid? do
+          errors
+        else
+          expected_options =
+            options
+            |> Enum.join(", ")
```
ZeLarpMaster, line 145: TIL that `Enum.join` with a list of atoms still works

test/support/fixtures.ex
```diff
@@ -51,6 +51,12 @@ defmodule Mobius.Fixtures do
 
     Socket.notify_payload(data, @shard)
 
+    # Yield to give the shard time to handle the message. Not a very safe way to
+    # do this, but since there's barely anything going on during tests it works
+    # well enough.
+    Process.sleep(5)
```
ZeLarpMaster, line 57: 5ms is not a long time, but I guess we'll see if the tests are unstable

ZeLarpMaster (changes requested):
Overall pretty good stuff, I like the way multipart is handled declaratively!

lib/mobius/actions/message.ex
Blond11516: Good point!

lib/mobius/validations/action_validations.ex
Blond11516: Yeah I initially did the conversion manually but I tried it out of curiosity and it worked 😛 Looking at the docs, anything "convertible to a string" works.

test/support/fixtures.ex
Blond11516: We really don't need a long delay here. I added it because I was getting one after some changes, but even with 1 ms I couldn't get a test to fail. I think that since there's basically nothing going on at that point just the fact that the process yields its thread is enough for the shard to pick up the message.

I bumped it up because I assumed the CI runners are slower than my machine, but I think it should work fine.

ZeLarpMaster (approved):

